PR #8, opened 2024-09-06
chore: update package name

commoddity:
Updates package name from `pokt-foundation/portal-middleware` to `buildwithgrove/path`.

### DIFF
--- a/config/morse/signed_aat.go
+++ b/config/morse/signed_aat.go
@@ -3,8 +3,8 @@ package morse
 import (
 	"fmt"
 
+	"github.com/buildwithgrove/path/config/utils"
 	sdk "github.com/pokt-foundation/pocket-go/provider"
-	"github.com/pokt-foundation/portal-middleware/config/utils"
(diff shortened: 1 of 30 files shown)


config/morse/signed_aat.go
```diff
@@ -3,8 +3,8 @@ package morse
 import (
 	"fmt"
 
+	"github.com/buildwithgrove/path/config/utils"
```
adshmh, line 6: Could you please update all imports so self-imports are at the bottom?

commoddity: Sure.

adshmh (changes requested):
Thank you for the PR, just one blocker.

config/morse/signed_aat.go
commoddity: > Could you please update all imports so self-imports are at the bottom?

Done.

adshmh (approved):

Issue #11, opened 2024-09-06
[Umbrella Ticket] PATH User Database 


Olshansk:
## Objective

Build the foundation for tracking users accessing the portal.

## Origin Document

[Postgres driver](https://github.com/pokt-foundation/portal-middleware/pull/760)
[Proposed Database Structure](https://www.notion.so/buildwithgrove/Proposed-Database-Migration-8c6980560a474e038d272a3198467a35)

## Goals

- Associated a relay with a user ID
- Enable other gateways to build features associated with a specific user
- Enable downstream efforts related to billing
- Enable downstream efforts related to rate limiting
- Enable downstream efforts for user-specific features in the future (whitelisting, user agents, settings, etc...)

## Deliverables

- [ ] Implementation that ties a `relay` to a `UserId` including
	- [ ] Connecting to, fetching and caching user data from a Postgres database.
	- [ ] Relating a service request to this user data via a UserApp ID (eg from .../v1/{USER_APP_ID} in the URL).
- [ ] User-specific rate limiting (dependent on #10 as a parallel requirement)
- [ ] Updated tooling and documentation to startup a postgres DB alongside the gateway
- [ ] 1-2 pager w/ docs to be turned in 🦖  docs in the future

## Non-goals / Non-deliverables

- Out-of-the box stripe integration or hooks
- Implementing downstream user settings (whitelisting, user agents, settings, etc...)
- Integrating with crypto payments

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk 
**Co-Owners**: @commoddity

fredteumer:
this feels like it's being conflated from a user db vs billing db situation. The purpose of the database is to codify the business  logic of the application for the purposes of user management, plan management, services, etc.

Billing, while important, inherits a small subset of this data and is combined with the metering of the application to integrate with Stripe.

commoddity:
> this feels like it's being conflated from a user db vs billing db situation. The purpose of the database is to codify the business logic of the application for the purposes of user management, plan management, services, etc.
> 
[...]

Issue #2, opened 2024-08-22
[CLI] Path MVP CLI


Olshansk:
## Objective

Create a `path` CLI foundation with MVP functionality to which we will add to over time.

## Origin Document

This discussion:

![Screenshot 2024-08-22 at 11 19 00 AM](https://github.com/user-attachments/assets/60de6549-84cd-4427-81e4-61d0fc05f5c6)

Along with this foundation in the `poktroll` repo:

![Screenshot 2024-08-22 at 11 19 14 AM](https://github.com/user-attachments/assets/c9e6b592-43d7-4721-b62b-de373d20139f)

## Goals

- Create the foundation for the `path` CLI

## Deliverables

- [ ] Identify which golang CLi framework we want to use; [spf13/cobra](https://github.com/spf13/cobra, [https://github.com/urfave/cli](urfave/cli)
- [ ] At a minimum, provide the following functionality:
	- [ ] `path gateway start`; starts the gateway process and will be used as the entrypoint from docker files
	- [ ] `path gateway scaffold`; walks the user through a series of questions and outputs a yaml file
	- [ ] `path test relay send`; a helper to send a test relay
- [ ] Provide short & long help messages 

## Non-goals / Non-deliverables

- Create a fully-fledged and documented CLI with all the features we need

## General deliverables

- [ ] **Comments**: Add/update TODOs and comments alongside the source code so it is easier to follow.
- [ ] **Testing**: Add new tests (unit and/or E2E) to the test suite.
- [ ] **Makefile**: Add new targets to the Makefile to make the new functionality easier to use.
- [ ] **Documentation**: Update architectural or development READMEs; use [mermaid](https://mermaid-js.github.io/mermaid/) diagrams where appropriate.

---

**Creator**: @Olshansk 
**Co-Owners**: @fredteumer @adshmh @commoddity

PR #15, opened 2024-09-09
feat: DB driver and cache

commoddity:
Related to https://github.com/buildwithgrove/path/issues/11

PR to add a Postgres database driver using SQLC to generate Go code from SQL files:
https://sqlc.dev/

This is the first prerequisite to allow associating a service request with a user for the purposes of metering, billing, rate limiting, user settings, etc.

This PR adds the following:
- `db/driver` package containing Postgres driver
- `db` package containing local cache of user data
- `user` package containing the `GatewayEndpoint` struct


db/driver/testdata/seed-test-db.sql
```diff
@@ -0,0 +1,19 @@
+INSERT INTO plans (type, rate_limit_throughput)
```
Olshansk, line 1: #PUC small comment at the top on what / how much this seeds.

db/driver/testdata/seed-test-db.sql
```diff
@@ -0,0 +1,19 @@
...
+    ('account_2', 'PLAN_UNLIMITED'),
+    ('account_3', 'PLAN_FREE');
+INSERT INTO user_apps (id, account_id, secret_key, secret_key_required)
+VALUES ('user_app_1', 'account_1', 'secret_key_1', TRUE),
+    ('user_app_2', 'account_2', 'secret_key_2', TRUE),
+    ('user_app_3', 'account_3', 'secret_key_3', TRUE),
+    ('user_app_4', 'account_1', NULL, FALSE),
+    ('user_app_5', 'account_2', NULL, FALSE);
+INSERT INTO user_app_allowlists (user_app_id, type, value)
+VALUES ('user_app_2', 'contracts', 'contract_1'),
```
Olshansk, line 15: What is this last set of values use dor?

config/config.go
```diff
@@ -73,7 +78,12 @@ func (c GatewayConfig) GetRouterConfig() RouterConfig {
 	return c.Router
 }
 
-func (c GatewayConfig) GetUserDataConfig() UserDataConfig {
+// UserDataEnabled returns true if user data handling is enabled for the Gateway.
+func (c GatewayConfig) UserDataEnabled() bool {
```
Olshansk, line 82:
```suggestion
func (c GatewayConfig) IsUserDataEnabled() bool {
```

config/config.go
```diff
@@ -126,6 +140,11 @@ func (c GatewayConfig) validate() error {
 	if err := c.validateServiceConfig(); err != nil {
 		return err
 	}
+	if c.UserData != nil {
```
Olshansk, line 143: Use the `IsUserDataEnabled` helper here

config/user_data.go
```diff
@@ -2,31 +2,19 @@ package config
...
+	"time"
 
 	"github.com/buildwithgrove/path/config/utils"
 )
 
+const defaultCacheRefreshInterval = 5 * time.Minute
+
 /* --------------------------------- User Data Config Struct -------------------------------- */
 
 // UserDataConfig contains user data configuration settings.
```
Olshansk, line 14: #PUC this comment. 

What's the goal of this structure from a functional perspective?

db/driver/driver.go
```diff
@@ -0,0 +1,87 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+
+	"github.com/buildwithgrove/path/user"
+)
+
+// The PostgresDriver struct satisfies the Driver interface which defines all database driver methods
```
Olshansk, line 13: Please enforce this via `var _  XXXInterface = &postgresDriver{}`

user/user_app.go
```diff
@@ -0,0 +1,68 @@
...
+	AllowlistTypeContracts  AllowlistType = "contracts"
+	AllowlistTypeMethods    AllowlistType = "methods"
+	AllowlistTypeOrigins    AllowlistType = "origins"
+	AllowlistTypeServices   AllowlistType = "services"
+	AllowlistTypeUserAgents AllowlistType = "user_agents"
+)
+
+type UserApp struct {
+	ID                  UserAppID
+	AccountID           AccountID
```
Olshansk, line 17: I'm getting kind of confused by all the terminology. Thinking of simplifying it to:

```
type UserApp struct {
	UserId UserId
	AppId AppId
```

Also, is `UserApp` the right term here or is it legacy?

For example, is `GatewayApp` more appropriate where we'd have many users attached to it?

user/user_app.go
```diff
@@ -0,0 +1,68 @@
...
+	PlanType            string
```
Olshansk, line 18: Thoughts on creating a type for plantype?

user/user_app.go
```diff
@@ -0,0 +1,68 @@
...
+	SecretKey           string
```
Olshansk, line 19: 1. What is this? #PUC
2. Should secrets be a separate struct?

user/user_app.go
```diff
@@ -0,0 +1,68 @@
...
+	SecretKeyRequired   bool
+	RateLimitThroughput int
+	Allowlists          map[AllowlistType]map[string]struct{}
```
Olshansk, line 22: What is this? #PUC or rename

README.md
```diff
@@ -185,3 +186,82 @@ make test_unit
...
+
+This is required for:
+
+- User specified app settings
+- Metering and billing of service requests
+- Rate limiting of service requests by throughput and/or capacity
+
+### Updated Endpoint
+
+Enabling user data will modify the endpoint for service requests to require a user app ID identifier at the end of the URL path.
```
Olshansk, line 204:
```suggestion
Enabling user data will modify the endpoint for service requests to require a user app identifier at the end of the URL path.
```

README.md
```diff
@@ -185,3 +186,82 @@ make test_unit
...
+
+For example:
+
+```bash
+http://eth-mainnet.localhost:3000/v1/{user_app_id}
```
Olshansk, line 209: Why do we call it a user_app instead of a gateway_app?

db/driver/sqlc/query.sql
```diff
@@ -0,0 +1,23 @@
+-- name: SelectUserApps :many
```
Olshansk, line 1: 1. Can you rename the filename from `query.sql` to something more semantic? E.g. `query_all_data.sql`
2. #PUC at the top on what / why we're using this for.

db/driver/sqlc/schema.sql
```diff
@@ -0,0 +1,42 @@
+CREATE TYPE rate_limit_capacity_period AS ENUM ('daily', 'weekly', 'monthly');
```
Olshansk, line 1: Why did you go with an enum for capacity but not throughput (e.g. per second, per minute, per hour)>

Trying to understand the thought process here

README.md
```diff
@@ -185,3 +186,82 @@ make test_unit
...
+  db_connection_string: "postgres://user:password@localhost:5432/database"
+```
+
+An example Postgres Docker configuration is included in the [docker-compose.yml](./docker-compose.yml) file at the root of this repository. **However, this configuration is not recommended for production use.**
+
+### Database Schema
+
+[Base Schema SQL File](./db/driver/sqlc/schema.sql)
+
+```mermaid
```
Olshansk, line 229: PTAL at my comments in `user/user_app.go`.

I'm trying to understand things like:

1. `UserApp` or `GatewayApp`?
2. Is the `id` in a `UserApp` a `UserId` or an `AppId`?
3. What is an `AccountId`? Is it the same as a `UserId`?
4. What's a `SecretKey`? Is it the `ApplicationPrivateKey` from a protocol perspective?

For all of these, we can hop into a discussion or feel free to rename it in case the questions I asked trigger some sort of thought process.

Olshansk (changes requested):
@commoddity Great start!

I did a high-level preliminary review (see screenshot below) but just wanted to pause it here because I have some questions w.r.t to the naming decisions.

Lmk if it makes sense or if we need to sync over a call.

Screenshot 2024-09-09 at 3 34 12 PM

user/user_app.go
```diff
@@ -0,0 +1,68 @@
...
+
+const (
+	AllowlistTypeContracts  AllowlistType = "contracts"
+	AllowlistTypeMethods    AllowlistType = "methods"
+	AllowlistTypeOrigins    AllowlistType = "origins"
+	AllowlistTypeServices   AllowlistType = "services"
+	AllowlistTypeUserAgents AllowlistType = "user_agents"
+)
+
+type UserApp struct {
```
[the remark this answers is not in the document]
commoddity: Have updated comments in `user_app.go` to clarify their function.

user/user_app.go
```diff
@@ -0,0 +1,68 @@
...
+	ID                  UserAppID
```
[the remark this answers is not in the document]
commoddity: The ID is the ID of the `UserApp` not the `User`. 

IMO `UserApp.ID` is clear. `UserApp.UserAppID` stutters, which is not ideal.

Makefile
```diff
@@ -59,3 +59,12 @@ copy_test_config: ## copies the example test configuration yaml file to .gitigno
...
 		echo ".config.test.yaml already exists, not overwriting."; \
 	fi
+
+#######################
+### SQLC Generation ###
+#######################
+
+.PHONY: sqlc_generate
+sqlc_generate: ## Generate SQLC code from db/driver/sqlc/*.sql files
+	sqlc generate -f ./db/driver/sqlc/sqlc.yaml
```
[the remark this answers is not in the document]
commoddity: Where should I add this? To the `README.md`?

db/driver/testdata/seed-test-db.sql
commoddity: Updated.

config/config.go
commoddity: Done.

config/user_data.go
commoddity: Just contains the DB connection string and an optional refresh interval (with default of 5 mins). Updated comment.

db/driver/driver.go
commoddity: Done. Also updated cache to accept driver as arg to avoid import cycle (this is a better design anyway).

user/user_app.go
commoddity: ID is not user ID though, it's the identifier for the user app itself. And Account ID is not the app ID.

See attached diagram for the relationship of entities in our current Gateway. [Will attach shortly]

user/user_app.go
commoddity: I like it and I have done it. :)

user/user_app.go
commoddity: Have updated comments in this file.

README.md
commoddity: A few thoughts on this:

1. An application in this context relates specifically to an end user (ie. someone using an endpoint, modifying settings, getting billed for relays, etc.) so I think the naming is appropriate.

2. Given the naming of the actors in Shannon I think having Gateways, Applications AND Gateway Applications could be confusing to most people.

That said, I typed this I had the thought that perhaps `Endpoint` could be a better name for this? WDYT?

db/driver/sqlc/query.sql
commoddity: The SQLC convention is to just have a single file called `query.sql` which contains all SQL queries used to autogenerate the Go code.In our case, there is only one query to get the User Apps.

I will update the comment to make this clear.

db/driver/sqlc/schema.sql
commoddity: I think throughput being defined by the second is the standard. Just my opinion here and I'm open to be wrong on this but I doubt anybody is going to want to be able to enforce throughput by units other than second. IMO no need to get that granular with it.

Whereas I see capacity as having more variability; people may want to bill/meter daily, weekly, monthly. For example we've now done both daily and monthly in our business.

README.md
commoddity: I've updated the comments in `user_app.go` to address these questions, but I will also answer them here.

At a high level, I am attempting to create a structure that will be roughly compatible with the current structure we use for account management and billing (note that what we call `PortalApp`s here are renamed to `UserApp`s):

Screenshot 2024-09-10 at 08 44 36

As you can see in the structure above, an Account may have multiple UserApps and also may have many Users (a team), with only one plan. Accounts are used for Plan and team management, while Apps store individual settings.

> * 1. `UserApp` or `GatewayApp`?
[See comment here](https://github.com/buildwithgrove/path/pull/15#discussion_r1751527366)
> * 2. Is the `id` in a `UserApp` a `UserId` or an `AppId`?
It's an app ID. It's what gets passed in the URL path, eg. `rpc.grove.city/v1/{USER_APP_ID}`
> * 3. What is an `AccountId`? Is it the same as a `UserId`?
Not the same as a user ID. One Account may have many users with different permissions and one Plan. 
> * What's a `SecretKey`? Is it the `ApplicationPrivateKey` from a protocol perspective?
No. A secret key is simply an optional authentication key. 

Might be worth renaming to `API key` or something similar to avoid confusion.

 It's passed in the headers as `Authentication: {SECRET_KEY}` (Note: perhaps we should allow it being passed as `Authentication: Bearer {SECRET_KEY}` as that is a common pattern.)

Hopefully this makes things clear @Olshansk . CC @adshmh as well for your thoughts.

user/user_app.go
commoddity: Comments have been updated.

Personally I would prefer not to create multiple structs. IMO the only purpose of the `UserApp` struct in PATH is to store settings like the secret key and so I think it makes semantic sense to have all fields at the base level. I prefer simplicity unless making things more complex has a clear benefit. Not strongly attached to this however.

README.md
Olshansk: Resolving this comment and will follow up in a new set of comments after review round #2.

db/driver/sqlc/schema.sql
Olshansk: > IMO no need to get that granular with it.

That's fine, just the type of thing worth calling out in comments as to why this decision was made.

user/user_app.go
```diff
@@ -0,0 +1,106 @@
+package user
+
+// A unique identifier for a user application. It must be passed as the last path segment of the service request URL.
```
Olshansk, line 3:
```suggestion
// A unique identifier for a user application.
// It must be passed as the last path segment of the service request URL like so:
// http://eth-mainnet.localhost:3000/v1/{user_app_id}
```

user/user_app.go
```diff
@@ -0,0 +1,106 @@
...
+	// must be passed as the `Authentication` header in service requests.
+	SecretKey string
+	// Whether the UserApp requires a secret key for authentication.
+	// If not set, the UserApp does not require a secret key for authentication.
+	SecretKeyRequired bool
+
+	// The allowlists for the UserApp, which allow the UserApp to specify certain security settings,
+	// for example, which contracts, methods, origins, services, or user agents are allowed to make
+	// service requests through the Gateway with the UserApp.
+	Allowlists map[AllowlistType]map[string]struct{}
```
Olshansk, line 60: Can we remove all of this (Auth/Secret Key, AllowList, etc...) from this PR?

Seems like there is a lot of discussion (naming, how we want to implement it, structure, etc...) that's outside the scope of this PR.

---

Rationale:
1. Auth is a solved problem. I want to understand if we should pick an opinionated solution (e.g. clerk, auth0, etc) and how we integrate it.
2. Allow/block lists is very critical but it looks like we're just tagging along poral-middleware techdebt. At the very least, this should be a separate file/structure/etc. 

Again, both of these can be multiple days of thought/review and I don't see the urgency in having a port of the portal-middleware solution here this week creating TECHDEBT so soon.

---

Note: As an engineer, if you're worried about the "but we already implemented it", what I would do is:
1. Remove it
2. `git diff . >> /tmp/tmp.diff`
3. Create a new GitHub issue and upload `tmp.diff`
4. Use that as a structure point in the followup

router/router_test.go
```diff
@@ -122,6 +122,8 @@ func Test_handleHTTPServiceRequest(t *testing.T) {
 	}
 }
 
+// TODO_TECHDEBT: replace all mockery autogenerated mocks with mockgen and move to own file
```
Olshansk, line 125:
```suggestion
// TODO_TECHDEBT(@commoddity): replace all mockery autogenerated mocks with mockgen and move to own file
```

README.md
```diff
@@ -185,3 +186,82 @@ make test_unit
...
+
+You may opt to enable user data config to unlock the ability to associate a specific user with a particular service request.
+
+This is required for:
+
+- User-specified app settings (e.g. secret key, allowlists, etc.)
+- Metering and billing of service requests (e.g. charging users $2 per 1 million requests)
+- Rate limiting of service requests by throughput (e.g. 30 req / second) and/or capacity (e.g. 1M req / month)
+
+### Updated Endpoint
```
Olshansk, line 202: I feel like whatever IDE you're using still doesn't have the "auto update table of contents" feature. The ToC should've automatically updated if you did.

README.md
Olshansk: I left some comments in `user_app.go` w.r.t the scope of this PR (i.e. removing auth and allow list data).

Also, per this explanation and the other comment, this is where my head is at right now (see the high level pseudo-struct below).

I'm not trying to over complicate, but trying to understand if this design is the right approach from first principles or just heavily influenced by how things are designed in portal-middleware today.

```go

type UserAccount {
	AccountId
	PlanType
	Auth
}

type GatewayEndpoint struct {
	// A unique ID specific to this endpoint
	EndpointID EndpointID

	UserAccount
	RateLimiting
}

type RateLimiting struct {
	ThroughputLimiting ...
	CapacityLimiting ...
}

type Auth struct {
	....
}
```

Olshansk (changes requested):
@commoddity Another partial review but stopping for now as I have some very high level thoughts & questions.

I anticipate a few more revisions on this PR and we might also need to get some thoughts / feedback from @adshmh

README.md
commoddity: > I left some comments in `user_app.go` w.r.t the scope of this PR (i.e. removing auth and allow list data).
> 
> Also, per this explanation and the other comment, this is where my head is at right now (see the high level pseudo-struct below).
> 
> I'm not trying to over complicate, but trying to understand if this design is the right approach from first principles or just heavily influenced by how things are designed in portal-middleware today.
> 
> ```go
> type UserAccount {
> 	AccountId
> 	PlanType
> 	Auth
> }
> 
> type GatewayEndpoint struct {
> 	// A unique ID specific to this endpoint
> 	EndpointID EndpointID
> 
> 	UserAccount
> 	RateLimiting
> }
> 
> type RateLimiting struct {
> 	ThroughputLimiting ...
> 	CapacityLimiting ...
> }
> 
> type Auth struct {
> 	....
> }
> ```

This looks fine except that I would suggest we keep Auth at the GatewayEndpoint (FKA UserApp) level.

config/user_data.go
```diff
@@ -2,31 +2,23 @@ package config
...
-	type temp UserDataConfig
-	var val struct {
-		temp `yaml:",inline"`
-	}
-	if err := value.Decode(&val); err != nil {
-		return err
-	}
-	*c = UserDataConfig(val.temp)
-	return c.validate()
+	DBConnectionString   string        `yaml:"db_connection_string"`
```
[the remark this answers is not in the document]
commoddity: I was thinking of renaming it to PostgresConnectionString to make it more clear since we only support one databaase.

user/user_app.go
commoddity: Regarding Auth I'm fine with moving it to (https://github.com/buildwithgrove/path/pull/18) since that's where it's actually used for the first time. That said, I don't really see the benefit to removing it only to put it back two PRs down. Your call.

> 1. Auth is a solved problem. I want to understand if we should pick an opinionated solution (e.g. clerk, auth0, etc) and how we integrate it.

I would strongly suggest we don't introduce an external dependency for this. As I mentioned above, it's really just a very simple API key check: does the request `Authorization` header contain the correct `SecretKey`? Yes or no. Adding something with auth0 is way overkill for this IMO.

_(Again, we can rename `SecretKey` to `APIKey` for clarity; I'm leaning towards this at this point.)_

> 2. Allow/block lists is very critical but it looks like we're just tagging along poral-middleware techdebt. At the very least, this should be a separate file/structure/etc.

This was actually my attempt to clean up/simplify the way we handle Allowlists in the the Portal. 

For example this is the struct we use currently:
```go
struct {
		Origins     map[Origin]struct{}                    `json:"origins,omitempty"`
		UserAgents  map[UserAgent]struct{}                 `json:"userAgents,omitempty"`
		Blockchains map[RelayChainID]struct{}              `json:"blockchains,omitempty"`
		Contracts   map[RelayChainID]map[Contract]struct{} `json:"contracts,omitempty"`
		Methods     map[RelayChainID]map[Method]struct{}   `json:"methods,omitempty"`
	}
```

That said, I'm fine with removing it for now if you like, but with the understanding that this is something we will be using very soon and it will need to be added back when we do.

README.md
commoddity: Sure, this is a great add-on. I've added to my IDE.

router/router_test.go
commoddity: This is already done in a subsequent PR which should be merged soon.

README.md
commoddity: I've updated the `UserApp` struct to `GatewayEndpoint`. I think it's an improvement for sure.

config/user_data.go
```diff
@@ -2,31 +2,23 @@ package config
...
+	CacheRefreshInterval time.Duration `yaml:"cache_refresh_interval"`
```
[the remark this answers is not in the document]
commoddity: Done.

config/user_data.go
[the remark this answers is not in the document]
commoddity: Updated.

config/user_data.go
```diff
@@ -2,31 +2,23 @@ package config
...
+const defaultCacheRefreshInterval = 5 * time.Minute
+
 /* --------------------------------- User Data Config Struct -------------------------------- */
 
-// UserDataConfig contains user data configuration settings.
+// UserDataConfig contains user data configuration settings, which are only relevant if user data handling
+// is enabled for the gateway by setting the 'user_data_config' field in the config YAML file.
+//
+// The DB connection string must be for a valid Postgres database, which will
+// contain user data for the Gateway. A cache refresh interval may also be set.
```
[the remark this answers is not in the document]
commoddity: Updated.

db/cache.go
```diff
@@ -0,0 +1,74 @@
...
+	userApps             map[user.UserAppID]user.UserApp
+	db                   Driver
+	cacheRefreshInterval time.Duration
+	mu                   sync.RWMutex
+	logger               polylog.Logger
+}
+
+type Driver interface {
+	GetUserApps(ctx context.Context) (map[user.UserAppID]user.UserApp, error)
+}
```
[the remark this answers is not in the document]
commoddity: Have updated the file.

user/user_app.go
commoddity: I've removed all references to `Allowlists` for now. We can discuss the implementation when we get to the point of adding them back.

I've kept `Auth`.

db/driver_mock_test.go
```diff
@@ -0,0 +1,56 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: driver.go
+//
+// Generated by this command:
+//
+//	mockgen -source=driver.go -destination=driver_mock_test.go -package=db
```
Olshansk, line 6: 1. Any changes we can put mocks in a `/mocks` package? `db/mocks/driver_mock.go`
2. Please add a makefile target that regenerates all mocks

db/postgres/driver.go
```diff
@@ -0,0 +1,98 @@
...
+		Queries: New(pool),
+		DB:      pool,
+	}
+
+	return driver, cleanup, nil
+}
+
+/* ---------- Query Funcs ---------- */
+
+func (d *postgresDriver) GetGatewayEndpoints(ctx context.Context) (map[user.EndpointID]user.GatewayEndpoint, error) {
```
Olshansk, line 64: #PUC gomoc (just short and useful)

db/postgres/driver.go
```diff
@@ -0,0 +1,98 @@
...
+	rows, err := d.Queries.SelectGatewayEndpoints(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	return d.convertToGatewayEndpoints(rows)
+}
+
+func (d *postgresDriver) convertToGatewayEndpoints(rows []SelectGatewayEndpointsRow) (map[user.EndpointID]user.GatewayEndpoint, error) {
```
Olshansk, line 73: #PUC gomoc (just short and useful)

db/postgres/driver.go
```diff
@@ -0,0 +1,98 @@
...
+	gatewayEndpoints := make(map[user.EndpointID]user.GatewayEndpoint, len(rows))
+
+	for _, row := range rows {
+		gatewayEndpoint := user.GatewayEndpoint{
```
Olshansk, line 77: Love how this is looking!

db/postgres/driver_test.go
```diff
@@ -0,0 +1,294 @@
...
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+				require.Equal(t, test.expected, endpoints)
+			}
+		})
+	}
+}
+
+/* -------------------- Dockertest Ephemeral DB Container Setup -------------------- */
```
Olshansk, line 204: Let's move this into a separate file

db/postgres/sqlc/schema.sql
```diff
@@ -0,0 +1,32 @@
...
+            AND rate_limit_capacity_period IS NOT NULL
+        )
+        OR (
+            rate_limit_capacity IS NULL
+            AND rate_limit_capacity_period IS NULL
+        )
+    )
+);
+CREATE TABLE user_accounts (
+    id VARCHAR(24) PRIMARY KEY,
```
Olshansk, line 24: Stupid question, is this the same as `account_id` or something else?

If so, can we give it the same name?

db/postgres/sqlc/schema.sql
```diff
@@ -0,0 +1,32 @@
+-- This file is used by SQLC to autogenerate the Go code needed by the database driver. 
+-- It contains all tables required for storing user data needed by the Gateway.
+-- See: https://docs.sqlc.dev/en/latest/tutorials/getting-started-postgresql.html#schema-and-queries
+--
+CREATE TYPE rate_limit_capacity_period AS ENUM ('daily', 'weekly', 'monthly');
+CREATE TABLE plans (
+    id SERIAL PRIMARY KEY,
```
Olshansk, line 7: Should/does this have a foreign key with `account_id` or `gateway_endpoints`?

Is there a row per plan type or for every plan that exists? If it's the type, sholdn't that be the primary key?

user/gateway_endpoint.go
```diff
@@ -0,0 +1,82 @@
...
+type UserAccount struct {
+	// The unique identifier for a UserAccount.
+	AccountID AccountID
+	// The plan type for a UserAccount, which identifies the pricing plan for the Account.
+	PlanType PlanType
+}
+
+// The rate limiting settings for a GatewayEndpoint.
+type RateLimiting struct {
+	// ThroughputLimit refers to rate limiting per-second (TPS).
```
Olshansk, line 53:
```suggestion
	// ThroughputLimit refers to rate limiting per-second (TPS).
	// This is used to prevent DoS or DDoS attacks.
```

user/gateway_endpoint.go
```diff
@@ -0,0 +1,82 @@
...
+	ThroughputLimit int
+	// CapacityLimit refers to rate limiting over longer periods, such as a day, week or month.
```
Olshansk, line 55:
```suggestion
	// CapacityLimit refers to rate limiting over longer periods, such as a day, week or month.
	// This is to prevent under billing or abuse of the services provided.
```

user/gateway_endpoint.go
```diff
@@ -0,0 +1,82 @@
...
+	CapacityLimit int
+	// The period over which the CapacityLimit is enforced. One of `daily`, `weekly` or `monthly`.
+	CapacityLimitPeriod CapacityLimitPeriod
+}
+
+// CapacityLimitPeriod is the period over which the capacity limit is enforced.
```
Olshansk, line 61: Great comment

user/gateway_endpoint.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (e *GatewayEndpoint) GetAuth() (string, bool) {
+	return e.Auth.APIKey, e.Auth.APIKeyRequired
+}
+
+func (e *GatewayEndpoint) GetThroughputLimit() int {
+	return e.RateLimiting.ThroughputLimit
+}
+
+func (e *GatewayEndpoint) GetCapacityLimit() (int, CapacityLimitPeriod) {
```
Olshansk, line 80: #PUC with godoc. I had to look in the code to understand.

Ditto above, but really no need to overomplicate it. I'm sure supermave or github copilot can even get the job done.

user/gateway_endpoint.go
```diff
@@ -0,0 +1,82 @@
...
+	// The authentication settings for the GatewayEndpoint.
+	Auth Auth
+	// The UserAccount that the GatewayEndpoint belongs to, including the PlanType.
+	UserAccount UserAccount
+	// The rate limiting settings for the GatewayEndpoint, which includes both
+	// the throughput (TPS) limit and the capacity (longer period) limit.
+	RateLimiting RateLimiting
+}
+
+// The authentication settings for a GatewayEndpoint.
```
Olshansk, line 32: nice

user/gateway_endpoint.go
```diff
@@ -0,0 +1,82 @@
+package user
```
Olshansk, line 1: love the shape this file took

README.md
```diff
@@ -185,3 +197,74 @@ make test_unit
...
+
+## 6. User Data
+
+By default, PATH does not associate user data with service requests (i.e. relays).
+
+You may opt to enable user data config to unlock the ability to associate a specific user with a particular service request.
+
+This is required for:
+
+- User-specified app settings (e.g. secret key, etc.)
```
Olshansk, line 209:
```suggestion
- User-specified app settings (e.g. api auth keys, etc.)
```

README.md
```diff
@@ -185,3 +197,74 @@ make test_unit
...
+- Rate limiting of service requests by throughput (e.g. 30 req / second) and/or capacity (e.g. 1M req / month)
+
+### 6.1 Updated Endpoint
+
+Enabling user data will modify the endpoint for service requests to require a gateway endpoint ID at the end of the URL path.
+
+For example:
+
+```bash
+http://eth-mainnet.localhost:3000/v1/{endpoint_id}
```
Olshansk, line 220:
```suggestion
http://eth-mainnet.localhost:3000/v1/{gateway_endpoint_id}
```

docker/docker-compose.full.yml
```diff
@@ -0,0 +1,41 @@
+services:
+  path_gateway:
+    container_name: path_gateway
+    build:
+      context: ..
+      args:
+        IMAGE_TAG: ${IMAGE_TAG:-development}
+    ports:
+      # TODO_IMPROVE: Update the port to match the actual value defined in the config yaml
```
Olshansk, line 9: Can you add more detail on what you mean here?

docker/docker-compose.full.yml
```diff
@@ -0,0 +1,41 @@
...
+    extra_hosts:
+      # Allows host machine to access the docker container; required if the
+      # full node RPC and gRPC URLs are running or port forwarded to localhost.
+      - "host.docker.internal:host-gateway"
+    volumes:
+      - ./cmd/.config.yaml:/app/.config.yaml
+    command: ["./path"]
+    depends_on:
+      - db
+    # TODO_IMPROVE: add a health check that looks for component ready states from the /healthz endpoint
```
Olshansk, line 20:
```suggestion
    # TODO_UPNEXT(#20): add a health check that looks for component ready states from the /healthz endpoint
```

docker/docker-compose.yml
```diff
@@ -2,9 +2,9 @@ services:
   path_gateway:
```
Olshansk, line 2: Why have two docker compose files? We can always up/down specific services regardless.

Olshansk (changes requested):
@commoddity REALLY love how this is shaping up. I think we're almost there and this should be my last round of a lot of comments.

We're close and ty for bearing with the iterations 🙏

README.md
```diff
@@ -5,26 +5,38 @@
...
+  - [4.1 Setup Config YAML](#41-setup-config-yaml)
+  - [4.2 Start the Container](#42-start-the-container)
+- [5. E2E Tests](#5-e2e-tests)
+  - [5.1 Running Tests](#51-running-tests)
+- [6. User Data](#6-user-data)
+  - [6.1 Updated Endpoint](#61-updated-endpoint)
+  - [6.2 Database Configuration](#62-database-configuration)
+  - [6.3 Database Schema](#63-database-schema)
+
+## 1. Introduction
```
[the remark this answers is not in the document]
commoddity: > ULTRA NIT: My recommendation is to avoid numbers. I've fallen down this path myself.
> 
> Whenever you edit or move things around, you end up needing to re-number everything.

I have no strong preference in this regard but the Markdown plugin you suggested actually has a feature that auto-numbers everything based on the H1, H2, H3, etc. tags. that's where the numbers comes from. Happy to remove though; your call.

config/config_test.go
```diff
@@ -67,8 +67,9 @@ func Test_LoadGatewayConfigFromYAML(t *testing.T) {
 					WriteTimeout:       5000 * time.Millisecond,
 					IdleTimeout:        5000 * time.Millisecond,
 				},
-				UserData: UserDataConfig{
-					DBConnectionString: "postgres://user:password@localhost:5432/database",
+				UserData: &UserDataConfig{
```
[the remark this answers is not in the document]
commoddity: I don't see the benefit. Database is only needed needed and used for user data so I think it makes things clearer to keep them all under that umbrella term.

eg. Is user data enabled? Then you need a database connection. Is it left off? No database needed. IMO this keeps it simple and clear.

config/user_data.go
```diff
@@ -2,38 +2,38 @@ package config
...
 )
 
+const defaultCacheRefreshInterval = 5 * time.Minute
+
 /* --------------------------------- User Data Config Struct -------------------------------- */
 
-// UserDataConfig contains user data configuration settings.
+// UserDataConfig contains user data configuration settings. This is off by default.
+//
+// If the 'user_data_config' field in the config YAML is set, the related features will become active.
```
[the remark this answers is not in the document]
commoddity: See above comment. Database requirement is strictly tied to user data functionality being enabled.

In the near future when we added throughput limiting there will be a requirement for other deps (eg. Redis, Gubernator sidecar, etc.) which are also under the user data umbrella. I think splitting all those deps out in their own config will be far more confusing to users than enforcing that everything required for user data lives under one umbrella term in the config.

db/cache.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/buildwithgrove/path/user"
+)
+
+// userDataCache is an in-memory cache that stores gateway endpoints and their associated data.
+type userDataCache struct {
+	db DBDriver
+
+	gatewayEndpoints     map[user.EndpointID]user.GatewayEndpoint
+	cacheRefreshInterval time.Duration
+	mu                   sync.RWMutex
```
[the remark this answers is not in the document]
commoddity: Sure I'll update this.

docker/docker-compose.full.yml
commoddity: The port is a configurable value so ideally we find a way to dynamically set the exposed port in the docker compose from the value set in the config YAML.

docker/docker-compose.yml
commoddity: My reasoning is that, with database enabled, the `path_gateway` container depends on the `db` container being up and healthy, ie. uses the `depends_on` property. Once we enable throughput limiting it will also depend upon at least one more dependency. 

To the best of my knowledge there is no way to selectively enable `depends_on`, meaning there are two less than ideal options:
1. Use `depends_on`, meaning that `path_gateway` will always run `db` et al, even when a user only wants to run `path_gateway` on its own (ie without user data/database).
2. Don't use `depends_on`. This means when we want to run all the deps we need to first start up on `db` container then manually wait an arbitrary time (eg. using `sleep _n_`) and then start up `path_gateway` to ensure PATH can connect to the DB on startup. This is doable but inelegant, requiring either a script or long awkward Make targets instead of the built-in `depends_on` solution provided by docker compose.

Using a separate docker compose allows both PATH on its own and PATh + user data to start up gracefully, ensuring that all PATH deps are running and healthy before the container itself runs.

db/postgres/sqlc/schema.sql
commoddity: It's the `type`. I've been advised against using strings as primary keys in the database. 

However AFAIK there is nothing wrong with doing do and I have done this many times before (currently the Portal DB `pay_plans` table uses a the pay plan type string as the PK).

I'm fine with making `type` the PK if you are.

user/user_app.go
commoddity: Done.

db/cache.go
```diff
@@ -0,0 +1,75 @@
...
+		<-time.After(c.cacheRefreshInterval)
+
+		err := c.setCache(ctx)
+		if err != nil {
+			c.logger.Error().Err(err).Msg("failed to refresh cache")
+		}
+	}
+}
+
+func (c *userDataCache) setCache(ctx context.Context) error {
```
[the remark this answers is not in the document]
commoddity: Done.

db/cache.go
```diff
@@ -0,0 +1,75 @@
...
+	gatewayEndpoints, err := c.db.GetGatewayEndpoints(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get gateway endpoints: %w", err)
+	}
+
+	c.mu.Lock()
+	c.gatewayEndpoints = gatewayEndpoints
+	c.mu.Unlock()
```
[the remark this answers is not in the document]
commoddity: Done.

db/cache.go
```diff
@@ -0,0 +1,75 @@
...
+
+func (c *userDataCache) GetGatewayEndpoint(ctx context.Context, endpointID user.EndpointID) (user.GatewayEndpoint, bool) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	gatewayEndpoint, ok := c.gatewayEndpoints[endpointID]
+	return gatewayEndpoint, ok
+}
+
+func (c *userDataCache) cacheRefreshHandler(ctx context.Context) {
```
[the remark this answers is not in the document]
commoddity: Done.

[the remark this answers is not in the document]
commoddity: Done.

db/driver_mock_test.go
commoddity: I believe Arash has had some objections to a common mocks package in the past so perhaps best to discuss with it.

db/postgres/driver.go
commoddity: Done.

user/gateway_endpoint.go
commoddity: Done.

user/gateway_endpoint.go
commoddity: Done.

user/gateway_endpoint.go
commoddity: Done.

docker/docker-compose.full.yml
commoddity: Updated comment.

docker/docker-compose.full.yml
commoddity: Done.

commoddity:
@Olshansk Have addressed or responded to all review comments. PTAL. We're almost there. :)

db/driver/driver.go
```diff
@@ -0,0 +1,97 @@
...
+	"fmt"
+
+	"github.com/jackc/pgx/v5"
+	"github.com/jackc/pgx/v5/pgxpool"
+
+	"github.com/buildwithgrove/path/db"
+	"github.com/buildwithgrove/path/user"
+)
+
+// The postgresDriver struct satisfies the db.Driver interface defined in the cache in the db package.
```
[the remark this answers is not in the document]
Olshansk: Sounds good to me. I can see that's it's already done so will resolve this comment and re-review.

README.md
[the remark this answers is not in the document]
Olshansk: As long as you're not spending manual time doing it, no objections.

config/config_test.go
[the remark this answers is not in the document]
Olshansk: Going to resolve and re-review with a clear perspetive.

config/user_data.go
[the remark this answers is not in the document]
Olshansk: Agreed, but I'm thinking from an outsider's perspective.

User: "How do I configure the database?"'
Us: "It's under user config"
User: "Why?"
Us: "Because the database is only used for user data"
User: "Oh okay, I get it".

_* Some time later *_

New user: "How do I configure the database?"
_Repeat conversation

_* Some time later *_

Us #1: "We need to reuse the database for something other than user"
Us #2: "Okay, do we need to split out the config?"
Us #1: "Probably, but that's a lot of changes right now"

db/driver_mock_test.go
Olshansk: @adshmh Do you mind chiming in on this topic?

While we don't have to resolve it here & now, I want to document the TODO (if it applies) w/ reasoning & rationale.

db/postgres/sqlc/schema.sql
Olshansk: Will re-review the data model to have a full understanding on it and will leave new comments.

docker/docker-compose.yml
Olshansk: I asked ChatGPT about this (see screenshot below) and it seems there is a `--no-deps` [1] option:

Screenshot 2024-09-16 at 11 12 13 AM

I like to bias to having less code (lower surface area for both maintenance and bugs, less cognitive overhead) and simply support this by having an extra make target.

[1] https://docs.docker.com/reference/cli/docker/compose/up/


Screenshot 2024-09-16 at 11 11 35 AM

config/user_data.go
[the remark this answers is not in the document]
Olshansk: Could you provide a quick tl;dr here for context? (Lots of different thread on different topics)

Olshansk (changes requested):
Not a review - just publishing my comments on previous threads and will dive into a full review shortly.

Makefile
```diff
@@ -59,3 +59,12 @@ copy_test_config: ## copies the example test configuration yaml file to .gitigno
 	else \
 		echo ".config.test.yaml already exists, not overwriting."; \
 	fi
+
```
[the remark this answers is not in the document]
commoddity: Left a comment on the other mention of this. I agree with having that but let's do it in a later PR. I'll add a TODO.

config/user_data.go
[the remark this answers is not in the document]
commoddity: tl;dr is with the Envoy Auth Plugin all authorization of requests is performed in the Go plugin that runs alongside the Envoy Proxy. This means that the database connection happens there, PATH doesn't know or care about a database and doesn't need anything related to either the database or user data.

[The full Notion doc has far more detail about this](https://www.notion.so/buildwithgrove/Envoy-Auth-Plugin-103a36edfff680aeb231d8bb01b4d4b8) but that's a quick summary.

docker/docker-compose.yml
commoddity: SGTM (and TIL). Not sure how I never came across the `no-deps` flag as I googled and ChatGPTed about it repeatedly. Now that I know I agree it's better to not have multiple docker compose files.

db/postgres/driver.go
```diff
@@ -0,0 +1,101 @@
...
+
+// convertToGatewayEndpoints converts a slice of the SelectGatewayEndpointsRow struct fetched from
+// the database to a map of the user.GatewayEndpoint struct that is used throughout the repo.
+func (d *postgresDriver) convertToGatewayEndpoints(rows []SelectGatewayEndpointsRow) (map[user.EndpointID]user.GatewayEndpoint, error) {
+	gatewayEndpoints := make(map[user.EndpointID]user.GatewayEndpoint, len(rows))
+
+	for _, row := range rows {
+		gatewayEndpoint := user.GatewayEndpoint{
+			EndpointID: user.EndpointID(row.ID),
+			Auth: user.Auth{
```
Olshansk, line 82: 1. Is this also going to be deprecated soon/quickly? 
2. Still unclear what we'll be using for auth. Is it a in-house solution?

Might be worth adding to the topic of discussions post review to get it over the finish line.

Olshansk (changes requested):

Olshansk:
@adshmh Would appreciate if you could review this as well.

router/router_test.go
commoddity: I've closed the router PR since the Envoy approach replaces the changes required in it so I've just made the router change in this branch rather than leaving a TODO.

config/user_data.go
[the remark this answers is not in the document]
commoddity: What do you think about just removing the user data/database from the config altogether since it's likely we will be moving the whole `db` package into the auth plugin. If something causes us not to take this approach I can add it back later with an improved name, though I think that's not likely to be necessary.

docker/docker-compose.yml
commoddity: Have updated to use only one `docker-compose.yml` and added separate make targets for PATh with & without auth components.

Makefile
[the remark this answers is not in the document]
commoddity: Added.

config/config_test.go
```diff
@@ -67,8 +67,9 @@ func Test_LoadGatewayConfigFromYAML(t *testing.T) {
...
+					PostgresConnectionString: "postgres://user:password@localhost:5432/database",
```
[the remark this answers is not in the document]
commoddity: IMO we should just remove the user data component from the config package altogether for now. If we need it later I can add it back but I think it's unlikely. See: https://github.com/buildwithgrove/path/pull/15#discussion_r1763411880

config/utils/utils.go
```diff
@@ -31,8 +31,8 @@ func IsValidURL(s string) bool {
 	return u.Scheme == "http" || u.Scheme == "https"
 }
 
-// IsValidDBConnectionString checks if a string is a valid PostgreSQL connection string.
-func IsValidDBConnectionString(s string) bool {
+// IsValidPostgresConnectionString checks if a string is a valid PostgreSQL connection string.
```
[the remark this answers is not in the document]
commoddity: Sure, that's fine.

docker/docker-compose.full.yml
```diff
@@ -0,0 +1,43 @@
...
+      - "3000:3000"
+    extra_hosts:
+      # Allows host machine to access the docker container; required if the
+      # full node RPC and gRPC URLs are running or port forwarded to localhost.
+      - "host.docker.internal:host-gateway"
+    volumes:
+      - ./cmd/.config.yaml:/app/.config.yaml
+    command: ["./path"]
+    depends_on:
+      - db
```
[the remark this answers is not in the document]
commoddity: Added.

db/cache.go
```diff
@@ -0,0 +1,75 @@
...
+	if err := cache.updateCache(context.Background()); err != nil {
+		return nil, fmt.Errorf("failed to set cache: %w", err)
+	}
+
+	go cache.cacheRefreshHandler(context.Background())
+
+	return cache, nil
+}
+
+func (c *userDataCache) GetGatewayEndpoint(ctx context.Context, endpointID user.EndpointID) (user.GatewayEndpoint, bool) {
```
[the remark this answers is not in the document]
commoddity: Added.

db/cache.go
```diff
@@ -0,0 +1,75 @@
...
+	db DBDriver
+
+	gatewayEndpoints     map[user.EndpointID]user.GatewayEndpoint
+	gatewayEndpointsMu   sync.RWMutex
+	cacheRefreshInterval time.Duration
+
+	logger polylog.Logger
+}
+
+func NewUserDataCache(driver DBDriver, cacheRefreshInterval time.Duration, logger polylog.Logger) (*userDataCache, error) {
```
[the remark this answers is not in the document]
commoddity: Added.

db/postgres/driver.go
commoddity: No, the entire `db` package will move to the `auth_plugin` module under the Envoy proposal.

db/postgres/sqlc/query.sql
```diff
@@ -0,0 +1,21 @@
...
+    p.rate_limit_capacity,
+    p.rate_limit_capacity_period
+FROM gateway_endpoints ge
+    LEFT JOIN user_accounts ua ON ge.account_id = ua.id
+    LEFT JOIN plans p ON ua.plan_type = p.type
+GROUP BY ge.id,
+    ua.plan_type,
+    p.rate_limit_throughput,
+    p.rate_limit_capacity,
+    p.rate_limit_capacity_period;
```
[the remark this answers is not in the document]
commoddity: Done.

Makefile
```diff
@@ -15,10 +15,18 @@ help: ## Prints all the targets in all the Makefiles
 
 .PHONY: path_up
```
Olshansk, line 16: Update to this whole section

```makefile
#############################
### Run Path Make Targets ###
#############################

.PHONY: path_up_gateway
path_up_gateway: ## Run just the PATH gateway without any dependencies
	docker compose up -d --no-deps path_gateway

.PHONY: path_up_build_gateway
path_up_build_gateway: ## Run and build just the PATH gateway without any dependencies
	docker compose up -d --build --no-deps path_gateway

.PHONY: path_up
path_up: ## Run the PATH gateway and all related dependencies
	docker compose up -d

.PHONY: path_up_build
path_up_build: ## Run and build the PATH gateway and all related dependencies
	docker compose up -d --build

.PHONY: path_down
path_down: ## Stop the PATH gateway and all related dependencies
	docker compose down
```

Makefile
```diff
@@ -59,3 +67,13 @@ copy_test_config: ## copies the example test configuration yaml file to .gitigno
...
+
+###############################
+### Generation Make Targets ###
+###############################
+
+.PHONY: sqlc_generate
+sqlc_generate: ## Generate SQLC code from db/driver/sqlc/*.sql files
+	sqlc generate -f ./db/driver/sqlc/sqlc.yaml
+
+# TODO_IMPROVE - add a make target for generating mocks
```
Olshansk, line 79:
```go
# // TODO_TECHDEBT(@commoddity): move all mocks to a shared mocks package
# // TODO_TECHDEBT(@commoddity): Add all other mock generation commands here
.PHONY: mocks_generate
mocks_generate: ## Generates mocks for all services used for tests
	TODO_IN_THIS_PR: JUST PUT ONE PLEASE
```

db/cache_test.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/buildwithgrove/path/user"
+)
+
+func Test_GetGatewayEndpoint(t *testing.T) {
+	tests := []struct {
+		name       string
+		endpointID user.EndpointID
+		mockReturn map[user.EndpointID]user.GatewayEndpoint
+		expected   user.GatewayEndpoint
+		found      bool
```
Olshansk, line 21: Reoder the params too - I'm aligning it with the order in which you're validating git.

```suggestion

		expectedEndpointFound      bool
		expectedGatewayEndpoint   user.GatewayEndpoint
```

db/cache_test.go
```diff
@@ -0,0 +1,173 @@
...
+			c.Equal(test.expected, gatewayEndpoint)
+		})
+	}
+}
+
+func Test_cacheRefreshHandler(t *testing.T) {
+	tests := []struct {
+		name       string
+		mockReturn map[user.EndpointID]user.GatewayEndpoint
+		expected   map[user.EndpointID]user.GatewayEndpoint
```
Olshansk, line 68: Ditto. In general, try to separate inputs and expected outputs (with better naming) so it's easier for other developers to understand.

Please update the tests too. A bit of a NIT but ty

cmd/main.go
```diff
@@ -85,12 +85,16 @@ func main() {
 		Relayer:           relayer,
 	}
 
-	apiRouter := router.NewRouter(gateway, config.GetRouterConfig(), logger)
+	apiRouter := router.NewRouter(router.RouterParams{
```
Olshansk, line 88: I don't think we should have `RouterParams`. Seems like an unnecessary level of indirection / abstraction.

Olshansk (changes requested):

config/user_data.go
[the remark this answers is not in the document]
commoddity: I've removed everything related to the database from the config.

config/config_test.go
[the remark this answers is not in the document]
commoddity: I have done this.

router/router.go
```diff
@@ -29,15 +29,21 @@ type (
 	}
 )
 
+type RouterParams struct {
```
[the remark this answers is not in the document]
commoddity: No. I've removed.

Makefile
commoddity: Done.

cmd/main.go
commoddity: Removed.

Olshansk (approved):
LGTM! Let's just get a 👍 from @adshmh as well before merging it in

adshmh (approved):
Pre-approving with one ask:

Please add TODOs to address the following:
- Our overall Authentication solution.
- Alternatives to DBs as source of truth: e.g. config files, API servers, etc.

Issue #12, opened 2024-09-06
[Infra] Mainline docker images for PATH 


Olshansk:
## Objective

Build and upload a docker image for every successful deployment to `main`.

## Origin Document

`ghcr` implementation in the poktroll repo: https://github.com/search?q=repo%3Apokt-network%2Fpoktroll+ghcr&type=code

## Goals

- Enable Grove and other gateways to pull a docker image rather than building one
- Enable downstream infrastructure efforts (e.g. scaling)

## Deliverables

- [ ] GitHub action to build a docker image on merges to main
- [ ] A link to the ghcr repo somewhere in the docs
- [ ] Tagging the first release!

---

**Creator**: @Olshansk 
**Co-Owners**: @HebertCL

HebertCL:
This is an ongoing effort related to improving Path's CI

PR #17, opened 2024-09-09
Infra: Docker images and status badges

HebertCL:
# TL;DR

Build workflow and README updates with badges for release and go mod.

## Summary

- Added `build-and-push.yml` workflow that builds and push Docker images to GitHub registry.
- Updated README mentioning the release section to reference the Docker images.
- Added README badges to reflect current Go mod version, build status (currently pointing to E2E for testing purposes) and release.

## Related issues

#12


.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,52 @@
+name: Publish image to GitHub Container Registry
+
+on:
+  push:
+    branches: ["main"]
+
+jobs:
+  publish-image:
+    runs-on: ubuntu-latest
```
commoddity, line 9: Can we lock this to a version, eg. 22.04

README.md
```diff
@@ -5,9 +5,16 @@
 </div>
 <br/>
 
+![GitHub Actions Workflow Status](https://img.shields.io/github/actions/workflow/status/buildwithgrove/path/run-e2e-tests.yml)
```
commoddity, line 8: Love this.

.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,52 @@
+name: Publish image to GitHub Container Registry
+
+on:
```
commoddity, line 3: I think this workflow should either run the tests itself or be triggered a successful run of the test workflow. WDYT?

.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,52 @@
...
+  publish-image:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout project
+        uses: actions/checkout@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Docker Metadata action
```
commoddity, line 17: Am I understanding this right that this tags the image based on the Github version tags we create manually on release? 

If so that seems sufficient for now rather than implementing automated versioning, although automating the semver tag creation is also an option.

commoddity (commented):
Great PR, just a few questions.

.github/workflows/build-and-push.yml
HebertCL: I agree with this. No sense in building an image from a commit where tests are failing. Question would be: Do we want to chain this to run right after e2e, or do we want it to build it only by running unit testing?

commoddity: My preference would be have a CI job that does all our checks: lint, unit test, E2E test, test Docker build, etc. then triggers this workflow if it all succeeds.

.github/workflows/build-and-push.yml
HebertCL: I merely ripped off this metadata config from poktroll repo. I haven't been able to see that your statement is what it does (like you, I assume that's how it works). I also think automating semver is an good option, so I would say whether we go for one or the other, we should be just fine.

.github/workflows/build-and-push.yml
HebertCL: Great catch. I'm fixing this.

.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      ## TODO: Is the GH token in place in the repo?
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ github.token }}
+
+      ## TODO: Do we want/need to make PATH multi arch?
```
Olshansk, line 43: Yes please, can you do it in this PR?

PATH is something I'd like to be able to run on my M1 processor.

Hit a lot of issues with in pocketdex so let's just get ahead of it: https://github.com/pokt-network/pocketdex/blob/1178ca47f68f6596bed7658dc1ba2106cdda1362/.github/workflows/ci-docker.yml#L33

cc @bryanchriswhite @okdas for visibility

.github/workflows/build-and-push.yml
```diff
@@ -0,0 +1,55 @@
...
+          images: |
+            ghcr.io/buildwithgrove/path
+          tags: |
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
+            type=ref,event=tag,suffix=-rc
+            type=sha,suffix=-rc
+            type=sha,format=long,suffix=-rc
+
+      ## TODO: Is the GH token in place in the repo?
```
Olshansk, line 35: This seems like a `TODO_IN_THIS_PR`.

See https://github.com/pokt-network/poktroll/blob/ecf74ced63cc920181827ebfb04bf03853fff210/.github/workflows/reviewdog.yml#L12

.github/workflows/run-e2e-tests.yml
```diff
@@ -2,12 +2,16 @@ name: Run E2E Tests
 
 on:
   push:
+    # paths-ignore:
```
Olshansk, line 5: remove?

Olshansk (commented):
Let's go with multi-arch and make sure we resolve the token issue in this PR.

Also, I usually use this to test github workflows locally: https://github.com/nektos/act

(Just in case you didn't know about it)

.github/workflows/build-and-push.yml
okdas: Yup. Also, more providers (including the lean services such as hetzner) add arm64 to their offerings, so I say we need it.

HebertCL: Added ARM arch for docker build.

.github/workflows/run-e2e-tests.yml
HebertCL: Removed.

.github/workflows/build-and-push.yml
HebertCL: I've solved this doubt thanks to @okdas

.github/workflows/build-and-push.yml
HebertCL: I've taken the workflows from `portal-middleware` to enhance this.

okdas (changes requested):
Looks great! I'm actually already using an image built by this PR in some load-testing work. ❤️ There are a couple of things I think we should do before we merge this.

1. `gocritic` is part of `golangci-lint`. You can enable it in `.golangci.yml` and remove the separate job (saves CI time):

```yaml
linters:
  disable-all: true
  enable:
    - gocritic
```
You can also change gocritic params that way: https://golangci-lint.run/usage/linters/#gocritic

2. It appears `golangci-lint` produces some errors in the CI. It would be great if they are addressed in the PR as well so they don't make it into other people's branches. Please let me know if you need assistance there.

.github/workflows/test-build-push.yml
```diff
@@ -0,0 +1,106 @@
...
+        with:
+          go-version: "1.22.3"
+
+      - name: Setup Git Authentication
+        run: |
+          git config --global url."https://${{ github.token }}:x-oauth-basic@github.com/".insteadOf "https://github.com/"
+
+      - name: Run unit tests
+        run: CGO_ENABLED=0 go test ./... -short
+
```
commoddity, line 61: Any reason we can't just run the E2E test job in this workflow and remove the separate file?

commoddity: OK fine with me. I generally prefer waiting a bit longer from more confidence from more comprehensive tests but I'm fine with keeping them separate. I can always just run the E2E locally.

.github/workflows/release.yml
```diff
@@ -0,0 +1,17 @@
+name: Tag and Release
+
+on:
+  push:
+    branches: ["main"]
```
commoddity, line 5: I think this should be triggered by a successful run of `Run tests, build image and push to GitHub Container Registry`.

.github/workflows/release.yml
```diff
@@ -0,0 +1,17 @@
...
+
+jobs:
+  release-on-push:
+    runs-on: ubuntu-22.04
```
Olshansk, line 9: @okdas For consistency, is this what you're using now too?

.github/workflows/test-build-push.yml
```diff
@@ -0,0 +1,125 @@
+name: Run tests, build image and push to GitHub Container Registry
```
Olshansk, line 1: Any chance we can follow the same pattern / split as in `poktroll`?

Specifically, we have a `main-build.yml`, a `release-artifacts.yml` a `run-tests.yml`.

I realize it's a little overkill, but don't think it should be too much overhead here.

https://github.com/pokt-network/poktroll/tree/main/.github/workflows

README.md
```diff
@@ -24,6 +35,10 @@
 
 It provides various tools and libraries to streamline the integration and interaction with decentralized protocols.
 
+## Path Releases
+
+Path releases provide a Docker image you can start using right away to bootstrap your Path gateway without the need of building your own image. Head over to [releases](https://github.com/buildwithgrove/path/releases) to download it.
```
Olshansk, line 40: Can you provide an example command on how to download the docker image?

#PUC (Please Update Comment)

Olshansk (changes requested):
Please re-reqest a review once it's ready.

.github/workflows/test-build-push.yml
HebertCL: I've arranged the jobs in a way where unit testing and e2e run in parallel, but the build job requires both of them to be successful before running the build. That would give us confidence that both types of tests get executed, taking less CI time and ensuring that any of them failing will not produce a non-functional image.

HebertCL:
> Looks great! I'm actually already using an image built by this PR in some load-testing work. ❤️ There are a couple of things I think we should do before we merge this.
> 
> 1. `gocritic` is part of `golangci-lint`. You can enable it in `.golangci.yml` and remove the separate job (saves CI time):
> 
> ```yaml
> linters:
>   disable-all: true
>   enable:
>     - gocritic
> ```
> 
> You can also change gocritic params that way: https://golangci-lint.run/usage/linters/#gocritic
> 
> 2. It appears `golangci-lint` produces some errors in the CI. It would be great if they are addressed in the PR as well so they don't make it into other people's branches. Please let me know if you need assistance there.

I've addressed this using a `golangci.yml` file that now the job is referring to for execution.

.github/workflows/release.yml
HebertCL: AFAICT poktroll is using `ubuntu-latest`. I agree with @commoddity tho that it is a better approach to fix the runner version. In this case, I'm using the same fixed version used in middleware.

See for reference:
https://github.com/pokt-network/poktroll/blob/3384138248474a2e724445384fbaff07ab0c41b2/.github/workflows/main-build.yml#L14
https://github.com/pokt-network/poktroll/blob/3384138248474a2e724445384fbaff07ab0c41b2/.github/workflows/release-artifacts.yml#L14

.github/workflows/test-build-push.yml
HebertCL: We can follow this pattern for the sake of consistency when it comes to files and naming convention. I'll take a look at the files and see what we can adapt (I've already taken parts of them).

.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,17 @@
+name: Tag and Release
+
+on:
```
commoddity, line 3: Shouldn't this one also only run when the tests pass?

.github/workflows/main-build.yml
```diff
@@ -0,0 +1,50 @@
+name: Build image and push to ghcr.io
+
+on:
+  workflow_run:
+    workflows: ["Path tests"]
+    types:
+      - completed
```
commoddity, line 7: I think this needs to specify a successful run, not just completed.

eg:
```yaml
on:
  workflow_run:
    workflows: ["Path tests"]
    types:
      - completed
    if: ${{ github.event.workflow_run.conclusion == 'success' }}
    ```

File: .github/workflows/run-tests.yml [hunk 15]
```diff
@@ -0,0 +1,81 @@
...
+          go-version: "1.22.3"
+
+      - name: Setup Git Authentication
+        run: |
+          git config --global url."https://${{ github.token }}:x-oauth-basic@github.com/".insteadOf "https://github.com/"
+
+      - name: Run golangci-lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          version: latest
```
commoddity, line 29: I wonder if we should peg this to a set version as well?

.github/workflows/main-build.yml
```diff
@@ -0,0 +1,50 @@
...
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ github.token }}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v5
+        with:
+          push: true
```
commoddity, line 44: Can we add the tag as a build arg?

eg.
```yaml
- name: Build and push Docker image
  uses: docker/build-push-action@v5
  with:
    push: true
    tags: ${{ steps.meta.outputs.tags }}
    build-args: IMAGE_TAG=${{ steps.meta.outputs.tags }}
....
```

commoddity:
> > Looks great! I'm actually already using an image built by this PR in some load-testing work. ❤️ There are a couple of things I think we should do before we merge this.
> > 
> > 1. `gocritic` is part of `golangci-lint`. You can enable it in `.golangci.yml` and remove the separate job (saves CI time):
> > 
> > ```yaml
> > linters:
> >   disable-all: true
> >   enable:
> >     - gocritic
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > You can also change gocritic params that way: https://golangci-lint.run/usage/linters/#gocritic
> > 
> > 2. It appears `golangci-lint` produces some errors in the CI. It would be great if they are addressed in the PR as well so they don't make it into other people's branches. Please let me know if you need assistance there.
> 
> I've addressed this using a `golangci.yml` file that now the job is referring to for execution.

Super minor NIT but there's a typo in the filename: it's `.golanci.yml` instead of `.golangci.yml`.

commoddity (changes requested):
Just a reminder on the comments I added; forgot to hit `Request Changes`.

.github/workflows/test-build-push.yml
HebertCL: Updated to reflect the suggested pattern

.github/workflows/release-artifacts.yml
HebertCL: This is true. We're pursuing to automate this release part and not tie it to a release we manually create. I'll update this workflow to reflect it.

README.md
HebertCL: This section has been updated.

.github/workflows/run-tests.yml
HebertCL: I think it's ok if we keep using `latest` since it's the golangci version and we're not pairing this with specific golang installation.

.github/workflows/main-build.yml
HebertCL: This has been addressed.

.github/workflows/release-artifacts.yml
HebertCL: This has been addressed.

HebertCL:
> > > Looks great! I'm actually already using an image built by this PR in some load-testing work. ❤️ There are a couple of things I think we should do before we merge this.
> > > 
> > > 1. `gocritic` is part of `golangci-lint`. You can enable it in `.golangci.yml` and remove the separate job (saves CI time):
> > > 
> > > ```yaml
> > > linters:
> > >   disable-all: true
> > >   enable:
> > >     - gocritic
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > You can also change gocritic params that way: https://golangci-lint.run/usage/linters/#gocritic
> > > 
> > > 2. It appears `golangci-lint` produces some errors in the CI. It would be great if they are addressed in the PR as well so they don't make it into other people's branches. Please let me know if you need assistance there.
> > 
> > 
> > I've addressed this using a `golangci.yml` file that now the job is referring to for execution.
> 
> Super minor NIT but there's a typo in the filename: it's `.golanci.yml` instead of `.golangci.yml`.

🤦🏻 no wonder my linter was failing. I've fixed this.

.github/workflows/main-build.yml
HebertCL: This has been addressed.

commoddity (approved):
LGTM

okdas (approved):
Looks great, thank you!

I've had issues with `on.workflow_run` in the past, and curious how that'll work with releases!

Olshansk (approved):

PR #21, opened 2024-09-10
fix: resolves panic when endpoint called for an unassigned service

commoddity:
This PR fixes a bug where the `gateway` package would panic due to `parser.GetHTTPErrorResponse` returning a nil value when a service request is sent for a service not configured in the config YAML.

### DIFF
--- a/request/errors.go
+++ b/request/errors.go
@@ -1,0 +1,29 @@
+package request
+
+import (
+	"fmt"
+	"net/http"
+)
+
+const (
+	// TODO_TECHDEBT: return the correct id field from the request
+	parserErrorTemplate = `{"jsonrpc":"2.0","id":"0","error":{"code":%d,"message":"%s"}}`
+)
+
+/* Parser Error Response */
+
+type ParserErrorResponse struct {
+	err string
+}
+
+func (r *ParserErrorResponse) GetPayload() []byte {
+	return []byte(fmt.Sprintf(parserErrorTemplate, http.StatusBadRequest, r.err))
+}
+
+func (r *ParserErrorResponse) GetHTTPStatusCode() int {
+	return http.StatusOK // JSON-RPC 2.0 spec requires a 200 status code even for errors.
+}
+
+func (r *ParserErrorResponse) GetHTTPHeaders() map[string]string {
+	return map[string]string{}
+}
--- a/request/errors.go
+++ b/request/errors.go
@@ -2,28 +2,25 @@ package request
 
 import (
 	"fmt"
-	"net/http"
 )
 
-const (
-	// TODO_TECHDEBT: return the correct id field from the request
-	parserErrorTemplate = `{"jsonrpc":"2.0","id":"0","error":{"code":%d,"message":"%s"}}`
-)
+const parserErrorTemplate = `{"code":%d,"message":"%s"}`
 
 /* Parser Error Response */
 
-type ParserErrorResponse struct {
-	err string
+type parserErrorResponse struct {
+	err  string
+	code int
 }
 
-func (r *ParserErrorResponse) GetPayload() []byte {
-	return []byte(fmt.Sprintf(parserErrorTemplate, http.StatusBadRequest, r.err))
+func (r *parserErrorResponse) GetPayload() []byte {
+	return []byte(fmt.Sprintf(parserErrorTemplate, r.code, r.err))
 }
 
-func (r *ParserErrorResponse) GetHTTPStatusCode() int {
-	return http.StatusOK // JSON-RPC 2.0 spec requires a 200 status code even for errors.
+func (r *parserErrorResponse) GetHTTPStatusCode() int {
+	return r.code
 }
 
-func (r *ParserErrorResponse) GetHTTPHeaders() map[string]string {
+func (r *parserErrorResponse) GetHTTPHeaders() map[string]string {
 	return map[string]string{}
 }
--- a/request/errors.go
+++ b/request/errors.go
@@ -1,10 +1,17 @@
 package request
 
 import (
+	"errors"
 	"fmt"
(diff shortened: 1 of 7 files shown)


request/parser.go
```diff
@@ -14,6 +14,13 @@ import (
 	"github.com/buildwithgrove/path/relayer"
 )
 
+const (
+	// TODO_DISCUSS: should we create a common errors package for all errors?
```
Olshansk, line 18: Yes, let's create an `request/errors.go` in this PR.

request/parser.go
```diff
@@ -14,6 +14,13 @@ import (
...
+	// TODO_IMPROVE: define specific error codes for all errors
+	parserErrorCode     = -32000
```
Olshansk, line 20: Can you give this a more concrete name like `evmExecutionReverted`? https://stackoverflow.com/questions/71296510/code-32000-message-execution-reverted

Not sure if it's correct, but just the first thing that came up

request/parser.go
```diff
@@ -84,3 +90,21 @@ func (p *Parser) getServiceID(host string) (relayer.ServiceID, error) {
 
 	return serviceID, nil
 }
+
+/* Parser Error Response */
```
Olshansk, line 94: I think this should go into `errors.go` as well.

Olshansk (changes requested):

request/parser.go
commoddity: [According to this it's:](https://ethereum-json-rpc.com/errors)
```
-32000
Invalid input
Missing or invalid parameters
```

But more specifically it seems there 32... error codes an EVM specific. So my question is do we want to be using this for an error the emerges before we even get to the point of selecting a specific service? Using 32... is a holdover from Middleware and I we agree that we should not bring techdebt from Middleware to PATH>

IMO we shouldn't and a simple http 400 status code should suffice here. WDYT?

Also CCing @adshmh for your thoughts on error codes in PATH generally.

request/parser.go
commoddity: Done.

request/parser.go
commoddity: Done.

commoddity:
@adshmh @Olshansk I've create an `errors.go` for the request package. PTAL so we are all aligned on the correct error code to return in this case (both in the HTTP response itself and the JSON-RPC response body).

Also note the TODO comments to return the correct request ID in the JSON-RPC body (once we have a proper method for unmarshalling JSON-RPC bodies to extract fields such as `id`).

request/errors.go
```diff
@@ -0,0 +1,29 @@
+package request
+
+import (
+	"fmt"
+	"net/http"
+)
+
+const (
+	// TODO_TECHDEBT: return the correct id field from the request
```
adshmh, line 9: This is out of scope for `request` package: it should not know about JSONRPC (or any other service spec). 
Could you please add a summary of the following to the PR, to help future readers/contributors?

```go
The responsibility of `request` package is: [Extract the service ID and find the target service's corresponding QoS instance.](https://github.com/buildwithgrove/path/blob/e0067eb0f9ab0956127c952980b09909a795b300/gateway/gateway.go#L52C2-L52C45).
Request package decides how the requested service is referenced by the user (currently: subdomain of the HTTP request). 
Processing should fail here only if:
A) It can not recognize the requested service, e.g. missing subdomain, or
B) The requested service is not found/configured for the gateway instance.

If things fail at this point: need to return a 404, because we do not understand what service the user is requesting.
```

adshmh (changes requested):
Thank you for the PR. 
The main blocker here is burdening the `request` package with the responsibilities of the `qos` package. No part of the code, other than the `qos` package should know what JSONRPC means.

commoddity:
> Thank you for the PR. The main blocker here is burdening the `request` package with the responsibilities of the `qos` package. No part of the code, other than the `qos` package should know what JSONRPC means.

@adshmh I've updated to remove refs to JSON RPC from `request`. Please take another look.

request/errors.go
commoddity: Added comment to `parser.go`

request/parser.go
commoddity: Outdated, resolving.

request/parser.go
```diff
@@ -1,10 +1,16 @@
-// Package request provides a struct for setting and retrieving relay
-// request details from the context during the relay request lifecycle.
+// The responsibility of the `request` package is to extract the service ID and find the target service's corresponding QoS instance.
+// See: https://github.com/buildwithgrove/path/blob/e0067eb0f9ab0956127c952980b09909a795b300/gateway/gateway.go#L52C2-L52C45
+//
+// Request package decides how the requested service is referenced by the user (currently: subdomain of the HTTP request).
+//
+// Processing should fail here only if:
+// A) No service is provided - Bad Request
+// B) The provided service is not found/configured for the gateway instance - Not Found
```
Olshansk, line 8: Let's make a small edit.

My main concern here is that it's very "what the requests package is for right now" rather than what it'll be for in the future. I've found that engineers never go back and update package level docs.

```suggestion
// Request package is responsible for parsing and forwarding user requests.
// It is not responsible for QoS, authorization, etc...
//
// For example, Processing should fail here only if:
// A) No service is provided - Bad Request
// B) The provided service is not found/configured for the gateway instance - Not Found
//
// The responsibility of the `request` package is to extract the service ID and find the target service's corresponding QoS instance.
// See: https://github.com/buildwithgrove/path/blob/e0067eb0f9ab0956127c952980b09909a795b300/gateway/gateway.go#L52C2-L52C45
```

Olshansk (approved):
Minor edit. Update the comment if necessary.

PR #22, opened 2024-09-12
feat: add scripts/quickstart.sh

commoddity:


Makefile
```diff
@@ -9,21 +9,33 @@ list: ## List all make targets
 help: ## Prints all the targets in all the Makefiles
 	@grep -h -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-60s\033[0m %s\n", $$1, $$2}'
 
+##############################
```
Olshansk, line 12: Seems like there's some overlap here with the other PRs so I'll review it afterwards.

commoddity: No rush on this one. Given the CLI work that can happen in parallel to PATH development I'm not even convinced we need to merge this into PATH itself. I think this concept (and logic) can likely be transferred to the CLI and just live there.

PR #27, opened 2024-09-19
Fix build workflow to trigger only for main branch

HebertCL:
# TL;DR

Fix for build workflow.

## Summary

This fix solves the issue with build workflow being triggered by pull requests. Instead, it changes the behavior to be triggered only by main branch.

## Related issues

#17

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -5,7 +5,8 @@ on:
     workflows: ["Path tests"]
     types:
       - completed
-    branches: ["main"]
+    branches: 
+      - main
 
 jobs:
   build-and-push:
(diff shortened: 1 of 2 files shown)


.github/workflows/main-build.yml
```diff
@@ -5,6 +5,7 @@ on:
     workflows: ["Path tests"]
     types:
       - completed
+    branches: ["main"]
```
commoddity, line 8: Need to add this to the release workflow as well.

Also can we use 

```yaml
branches:
    - main
    ```
    syntax for consistency?

Comment (commoddity) [inline .github/workflows/main-build.yml:8]: Need to add this to the release workflow as well.

Also can we use 

```yaml
branches:
    - main
```
    syntax for consistency?

commoddity (changes requested):

commoddity (approved):
LGTM

PR #34, opened 2024-09-25
[Docker] Update the tag of the built and published docker images

Olshansk:
Two main changes:
- Always tag the latest image with `latest`
- Use a shortened version of the `sha`

Where images are stored: https://github.com/buildwithgrove/path/pkgs/container/path

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -31,7 +31,6 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=ref,event=tag,suffix=-rc
-            type=sha,suffix=-rc,format=short
             type=sha,format=short,suffix=-rc
             type=ref,event=branch,pattern=latest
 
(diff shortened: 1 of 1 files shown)


.github/workflows/main-build.yml
```diff
@@ -31,8 +31,9 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=ref,event=tag,suffix=-rc
-            type=sha,suffix=-rc
-            type=sha,format=long,suffix=-rc
+            type=sha,suffix=-rc,format=short
+            type=sha,format=short,suffix=-rc
```
HebertCL, line 35: This line seems to contain the same parameters as previous one, just different order. Is that intended?

HebertCL (changes requested):
Just a NIT on what seems an unnecessary duplicated line

HebertCL (approved):
LGTM

Olshansk:
Screenshot 2024-09-30 at 4 30 13 PM

PR #39, opened 2024-10-03
[QoS] Support publishing local (single instance) PATH QoS data

adshmh:
This enables publishing and usage of QoS data (from both user requests and the `EndpointHydrator`) in a single PATH instance. This enables running a single PATH instance with the improved user success rate from the existing QoS implementation, without introducing any extra dependencies.


cmd/main.go
```diff
@@ -36,26 +35,49 @@ func main() {
...
+		QoSPublisher:      qosPublisher,
+		Logger:            logger,
 	}
 
 	// Until all components are ready, the `/healthz` endpoint will return a 503 Service
 	// Unavailable status; once all components are ready, it will return a 200 OK status.
 	// health check components must implement the health.Check interface
 	// to be able to signal they are ready to service requests.
+	components := []health.Check{protocol}
+	if hydrator != nil {
```
commoddity, line 75: Nice, love the usage of this.

cmd/message.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/buildwithgrove/path/gateway"
+	"github.com/buildwithgrove/path/message/local"
+)
+
+// getQoSPublisher returns an instance of publisher
+// which meets the requirements of the gateway package
+// for QoSPublisher.
+// The returned publisher instance is determined by the
+// messaging system's config.
+// e.g. return a "local" publisher if no messaging systems
```
commoddity, line 14: Excellent, this will be useful for us doing local dev.

cmd/qos.go
```diff
@@ -0,0 +1,74 @@
...
+// the gateway packages' QoSEndpointCheckGenerator interface.
+func getServiceQoSInstances(
+	gatewayConfig config.GatewayConfig,
+	logger polylog.Logger,
+) (
+	map[relayer.ServiceID]gateway.QoSService,
+	map[relayer.ServiceID]gateway.QoSEndpointCheckGenerator,
+	error,
+) {
+	// Build a mpa of services configured for the hydrator
```
commoddity, line 27: NIT: typo

cmd/qos.go
```diff
@@ -0,0 +1,74 @@
...
+	gatewayServiceIDsIdx := buildServiceIDsIdx(gatewayConfig.GetEnabledServiceIDs())
+
+	// TODO_TECHDEBT: refactor this function to remove
+	// the need for manually adding entries for every new QoS implmenetation.
+	gatewayQoSService := make(map[relayer.ServiceID]gateway.QoSService)
+	hydratorQoSGenerators := make(map[relayer.ServiceID]gateway.QoSEndpointCheckGenerator)
+
+	// TODO_FUTURE: support serviceQoS-specific configuration.
+	allServiceIDs := append(gatewayConfig.GetEnabledServiceIDs(), gatewayConfig.HydratorConfig.ServiceIDs...)
+	for _, serviceID := range allServiceIDs {
```
commoddity, line 39: Should we move this to its own file?

config/service.go
```diff
@@ -0,0 +1,13 @@
+package config
+
+import (
+	"github.com/buildwithgrove/path/relayer"
+)
+
+const (
```
commoddity, line 7: Nice, I prefer this implementation. Simplifies the config YAML and makes it easy to find which service uses which qos.

health/checker.go
```diff
@@ -35,10 +35,12 @@ type (
 		Components []Check
 		Logger     polylog.Logger
 	}
+
 	// health.Check is an interface that must be implemented
 	// by components that need to report their health status
 	Check interface {
-		Name() string  // Name returns the name of the component being checked.
+		Name() string // Name returns the name of the component being checked.
+		// TODO_FUTURE: consider adding a message/reason for an unhealthy status.
```
commoddity, line 43: Agreed, this will be a good improvement.

message/local/local.go
```diff
@@ -0,0 +1,29 @@
+// package local provides the publishing functionality
+// required to skip sharing data between PATH instances,
+// by informing the local PATH components instead.
+// This allows the running of a single PATH instance
+// without the need for setting up a messaging platform,
+// e.g. NATS, REDIS.
```
commoddity, line 6: This is great.

commoddity (approved):
This looks great. I will dig deeper into the full QoS implementation to get caught up but I see no reason to holdup approving this PR now. Approved.

cmd/qos.go
adshmh: Thanks, fixed.

cmd/qos.go
adshmh: I agree, but for the code being introduced in this PR I think this is sufficient.
We will likely split the code into more files once we address some of the TODOs here.

cmd/qos.go
```diff
@@ -0,0 +1,74 @@
...
+		switch serviceID {
+		case config.ServiceIDEVM:
```
Olshansk, line 41: It's unclear to me why we'd have different implementations here. 

#PUC?

Issue #42, opened 2024-10-10
[Documentation] Add `docusaurus` documentation


Olshansk:
## Objective

Add modern-looking docs to the `path` repo

## Origin Document

[docusaurus.io](https://docusaurus.io) was used in the [poktroll repo here](https://github.com/pokt-network/poktroll/tree/main/docusaurus) to deploy the documentation at [dev.poktroll.com](https://dev.poktroll.com).

<img width="975" alt="Screenshot 2024-10-10 at 11 23 30 AM" src="https://github.com/user-attachments/assets/031b2e2d-d51c-42ff-aab3-2251e45d6c9d">

## Goals

- Make PATH onboarding simple, easy & attractive w/ good, easy-to-navigate and visually asthetic documentation

## Deliverables

- [ ] [@Olshansk] Add docusarusus to the docs in the same way it's implemented in poktroll
- [ ] [@Olshansk] Link to 👇 in the README
- [ ] [@Olshansk] Add Makefile helpers (see image above) in the `Makefile` of this repo for development purposes
- [ ] [@okdas] Deploy the docs to `docs.grove.city` similar to `dev.poktroll.com`
- [ ] [@fredteumer] Link to ☝️ somewhere on grove.city

## Non-goals / Non-deliverables

- Updating the existing documentation
- Adding new documentation

---

**Creator**: @Olshansk 
**Co-Owners**: @fredteumer

PR #55, opened 2024-10-31
Create initial setup and workflow for docusaurus

HebertCL:
# TL;DR

Create initial Docusaurus setup to handle static page documentation.

## Summary

This PR handles the following components:

- Docusaurus folder structure and basic page for Path static page documentation.
- GitHub Actions workflow to handle the deployment of documentation static page.

## TODO

- [x] Setup GitHub Pages to mirror `poktroll` configuration.
- [x] Setup `path.grove.city` DNS recordset for the page.
- [x] Enable pipeline to generate the documentation automatically.

## Related issues

#42


docusaurus/README.md
```diff
@@ -0,0 +1,41 @@
+# Website
```
Olshansk, line 1: Is this something you added or was it a default?

I'm thinking we just update it to use `make docusaurus_start` from the main Makefile. Wdyt?

Olshansk (changes requested):
@HebertCL Thanks for kicking this off!

1. PTAL at admonitions. It's the best way to add notes in the future: https://docusaurus.io/docs/markdown-features/admonitions

2. I updated the main README to use ☝️ as a reference.

3. Can you add an empty `develop` and `operate` section similar to what we have in dev.poktroll.com.

It took me a little while to figure it out the first time, but just copy-pasting the pattern we have there should be enough.

**Why now?** If you set it up/configure it, it'll incentivize the rest of the team to actually add docs :)

Screenshot 2024-11-07 at 2 08 33 PM

HebertCL:
> 3. Can you add an empty `develop` and `operate` section similar to what we have in dev.poktroll.com.

I've taken care of this and confirm that the validation is passing and building is working correctly in local. I've went ahead and request re-review.

Olshansk (approved):
Let's :shipit: 

Thanks @HebertCL!

PR #74, opened 2024-12-03
[READY] Localnet configuration parity with staging

HebertCL:
# TL;DR

Configuration parity with staging and localnet.

## Summary

Added path-values file and descriptions to spin up and tear down localnet. Values file provides parity with the current staging deployment using same chart.

Additionally updated README to describe steps to bring up/down localnet and Makefile targets for it.


commoddity (commented):
LGTM but just curious if you can give some more context on why we need to use Kind in addition to Tilt?

README.md
```diff
@@ -288,9 +291,29 @@ make test_unit
...
+## 8. Running Localnet
 
-### 8.1. Docker Permissions Issues - Need to run sudo?
+You can use path configuration under `/local` to spin up a local development environment using `Kind` + `Tilt`.
+
+### 8.1 Pre-requisites
+
+Make sure you have the following tools installed:
+
+- [Kind](https://kind.sigs.k8s.io/#installation-and-usage)
```
Olshansk, line 302: We have a pre-requisites section in `1.1`. Can you move it there?

Makefile
```diff
@@ -194,6 +194,26 @@ config_morse_localnet: ## Create a localnet config file to serve as a Morse gate
 		echo "##################################################################################################################"; \
 	fi
```
Olshansk, line 196: I looked at the top of this Makefile and it seems like it's still using `docker compose`. 

Please merge & consolidate with `main`.

local/path/config/path-values.yaml
```diff
@@ -0,0 +1,46 @@
+global:
```
Olshansk, line 1: Why is this not part of the `helm-charts` repo?

README.md
```diff
@@ -288,9 +291,29 @@ make test_unit
 make test_e2e_shannon_relay
 ```
 
-## 8. Troubleshooting
+## 8. Running Localnet
 
-### 8.1. Docker Permissions Issues - Need to run sudo?
+You can use path configuration under `/local` to spin up a local development environment using `Kind` + `Tilt`.
```
Olshansk, line 296: See what we have in [1]

Once the user has gotten to step `8`, they don't want to relearn pre-requisites again, they just want to do that all upfront.

https://dev.poktroll.com/develop/developer_guide/quickstart.

Makefile
```diff
@@ -194,6 +194,26 @@ config_morse_localnet: ## Create a localnet config file to serve as a Morse gate
...
 
+###############################
+###  Localnet Make targets  ###
+###############################
+.PHONY: localnet_up
```
Olshansk, line 200:
```suggestion

.PHONY: localnet_up
```

Makefile
```diff
@@ -194,6 +194,26 @@ config_morse_localnet: ## Create a localnet config file to serve as a Morse gate
...
+localnet_up: ## Spins up Kind cluster for local development and brings up Tilt from file
```
Olshansk, line 201: Any chance you can make this similar to what we have in `poktroll` for consistency?

```Makefile
.PHONY: localnet_up
localnet_up: check_docker_ps check_kind_context proto_regen localnet_regenesis ## Starts up a clean localnet
	tilt up
```

- Avoid inline flags
- Assume the cluster is already created
- Etc...

A couple other high level considerations:
- You can create a target called `make kind_config`  to do the things you're trying
- After you merge with main, check if there are any other better consolidations to be had.

README.md
Olshansk, line 296: Let's add the same explanation of why we need kind

cc @commoddity 

Screenshot 2024-12-03 at 7 45 43 PM

Olshansk (changes requested):
@HebertCL GOing to take another look after the chances are addressed. 

1. Make sure to merge with main
2. High-level meta piece of feedback: try to approach the docs holistically so it seems like one whole wrote the whole thing.

It's not easy, but it goes a long way.

In this specific instance, the actionable is to consolidate pre-requisite secionts.

README.md
HebertCL: Moved the pre-requisites to deployment section.

Makefile
HebertCL: I pulled main and verified the changes in main. The docker-compose section is still [there](https://github.com/buildwithgrove/path/blob/main/Makefile#L14-L28). I will remove those targets as well as the tilt ones in favor of the configuration I am pushing in this PR.

local/path/config/path-values.yaml
HebertCL: It IS part of the Path helm chart. 

The global directive encapsulates configurations that may spawn multiple times and across different resources. You can find references in [deployment](https://github.com/buildwithgrove/helm-charts/blob/main/charts/path/templates/deployment.yaml) and [serviceaccount](https://github.com/buildwithgrove/helm-charts/blob/main/charts/path/templates/serviceaccount.yaml) resources.

README.md
HebertCL: I've moved the pre-requisites to 1.1

Makefile
HebertCL: This has been addressed.

Olshansk:
@HebertCL Please re-request a review when this is ready.

Makefile
HebertCL: I have split into `localnet.mk` to handle everything localnet-related.

README.md
HebertCL: Added a comment explaining the potential situation when using K8s Docker desktop.

Olshansk (approved):

Olshansk:
@HebertCL I pushed a few changes but otherwise LGTM!

PR #79, opened 2024-12-04
[Logging] Add logging in the HTTP request processing path

adshmh:
Add logging for all cases of processing HTTP requests in the gateway.


commoddity (approved):
LGTM, this will add some valuable data to the logs.

PR #85, opened 2024-12-05
Update Shannon E2E config with Shannon Beta Testnet URLs

adshmh:
Update Shannon Example Config for E2E tests with Shannon Beta Testnet URLs.


commoddity (approved):
LGTM

PR #95, opened 2024-12-11
Fix: make the /v1 endpoint more permissive

adshmh:


commoddity (approved):
LGTM. I will experiment with this as part of the full Envoy Proxy stack and see if any modifications are required there.

Olshansk (approved):

PR #99, opened 2024-12-12
fix: set lazy mode to true when setting shannon e2e test config

commoddity:
This is to minimize the confusing log entries if an E2E test fails.

Discord conversation:
https://discord.com/channels/824324475256438814/1316586048670666802/1316744288910184560

### DIFF
--- a/e2e/scripts/update_shannon_config_from_secrets.sh
+++ b/e2e/scripts/update_shannon_config_from_secrets.sh
@@ -18,9 +18,12 @@ update_shannon_config_from_env() {
         return 1
     fi
 
+    # Set the config values from the Github secrets environment variables.
+    # This also sets lazy_mode to minimize confusing log entries if an E2E test fails in the CI.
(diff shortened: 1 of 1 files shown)


e2e/scripts/update_shannon_config_from_secrets.sh
```diff
@@ -18,9 +18,12 @@ update_shannon_config_from_env() {
         return 1
     fi
 
+    # Set the config values from the Github secrets environment variables.
+    # This also sets lazy_mode to minimize confusing log entries if an E2E test fails in the CI.
```
Olshansk, line 22:
```suggestion
    # Update the PATH Shannon config to reflect secrets on GitHub.
    # Also set lazy_mode to true in order to minimize verbosity of log entries in CI E2E tests.
```

Olshansk (approved):
@commoddity One comment but otherwise LGTM.

PR #102, opened 2024-12-13
[ENVOY] feat: Envoy to set Service ID header from subdomain

commoddity:
This PR makes three main changes to PATH:

1. A new Envoy Proxy HTTP Filter to Extract the Service ID from request `subdomain` or `header`
2. Removes the concept of a service `"alias"` from PATH's code; instead all aliases are handled at the Envoy level.
3. Introduces a map in `.allowed-services.lua` to configure which services are allowed by a PATH instance.

_Further details below._

1. **Envoy Lua Filter**
Adds a new Lua HTTP filter to the Envoy YAML template to set the `target-service-id` header from the subdomain or the header _(including resolving service aliases to authoritative service IDs - see point 2.)_.

`"anvil.gateway.pokt.network" -> Header: "target-service-id: anvil"`
`"eth.gateway.pokt.network" -> Header: "target-service-id: F00C"`

This allows us to continue supporting the legacy method of passing chain IDs to middleware via the subdomain.

eg.
```bash
curl http://localhost:3001/v1 \
    -X POST \
    -H "Content-Type: application/json" \
    -H "Endpoint-ID: 1a2b3c4d" \
    -H "Authorization: <API_KEY>" \
    -H "Target-Service-ID: anvil" \
    -d '{"jsonrpc": "2.0", "id": 1, "method": "eth_blockNumber" }'
```
and
```bash
curl http://anvil.localhost:3001/v1 \  
    -X POST \
    -H "Content-Type: application/json" \
    -H "Endpoint-ID: 1a2b3c4d" \
    -H "Authorization: <API_KEY>" \
    -d '{"jsonrpc": "2.0", "id": 1, "method": "eth_blockNumber" }'
```
both work when hitting Envoy _(in the latter example, the new Envoy HTTP filter sets the `Target-Service-ID` header from `anvil.localhost:3001`_). 

2. **Removal of service `"alias"` concept from PATH**

_tl;dr all service IDs forwarded to PATH from Envoy are 1. in the `target-service-id` header and 2. authoritative IDs, not aliases_

All handling of service aliases requests is done at the Envoy Proxy level now, inside the Lua filter.

Allowed services for a PATH instance are now configured inside a new `.allowed-services.lua` file which is mounted into the Envoy Proxy container and read from inside the Lua filter.

If requests are made for a service that does not exist in this file they will be rejected.

Aliases are also configured here and the Lua filter ensures that requests made for aliases are forwarded to PATH with the correct `authoritative service ID` set as the `target-service-id` header.

_NOTE: This PR also updates all header string consts in the Go code to be lowercase as the variable casing was causing issues with Envoy. Suggest that all header consts defined in the Go code be lower case going forward to avoid the potential for casing issues with Envoy. It also removes the deprecated `x-` prefix from header names._


envoy/auth_server/auth/endpoint_id_extractor.go
```diff
@@ -69,12 +69,14 @@ var _ EndpointIDExtractor = &HeaderExtractor{}
...
 // ExtractGatewayEndpointID extracts the endpoint ID from the HTTP headers.
-// The endpoint ID is expected to be in the "x-endpoint-id" header.
+// The endpoint ID is expected to be in the "endpoint-id" header.
 //
-// eg. Header = "x-endpoint-id: 1a2b3c4d" -> endpointID = "1a2b3c4d"
+// eg. Header = "endpoint-id: 1a2b3c4d" -> endpointID = "1a2b3c4d"
 func (h *HeaderExtractor) extractGatewayEndpointID(req *envoy_auth.AttributeContext_HttpRequest) (string, error) {
 	headers := req.GetHeaders()
 
+	fmt.Println("headers", headers)
```
Olshansk, line 78: Remove?

Olshansk (changes requested):
LGTM! 

Will approve after we merge the base and rebase this.

commoddity:
> LGTM!
> 
> Will approve after we merge the base and rebase this.

Thank you! Have updated from `main` and fixed merge conflicts.

Olshansk (commented):
@commoddity The `introduction.md` is great, and I know you're also working on a cheatsheet, but this document is getting to the point that only we (core team) would ever go through it from start to finish.

Can you add a `TODO_DOCUMENT(@commoddity): Document all envoy configs similar to https://dev.poktroll.com/operate/configs/relayminer_config and work on that before the EOW?

I'll keep reviewing this tomorrow.

commoddity:
> @commoddity The `introduction.md` is great, and I know you're also working on a cheatsheet, but this document is getting to the point that only we (core team) would ever go through it from start to finish.
> 
> Can you add a `TODO_DOCUMENT(@commoddity): Document all envoy configs similar to https://dev.poktroll.com/operate/configs/relayminer_config and work on that before the EOW?

Yes definitely. I share the same concern about this doc getting too big for its boots; I think a shorter description of the config required to run Envoy is a good idea.

Makefile
```diff
@@ -235,33 +235,13 @@ init_envoy: copy_envoy_config copy_gateway_endpoints ## Runs copy_envoy_config a
```
Olshansk, line 235: @commoddity Can you please update the suffixes on endpoints 1,2,3 in `envoy/auth_server/endpoint_store/endpoint_store_test.go`?

envoy/allowed-services.template.lua
```diff
@@ -0,0 +1,147 @@
...
+-- The service IDs configured here are used in the `envoy.filters.http.lua` HTTP filter defined in `.envoy.yaml` config file.
+-- The `.allowed-services.lua` file must be mounted as a file in the Envoy Proxy container at `/etc/envoy/.allowed-services.lua`.
+--
+-- If you wish to define aliases for existing services, you must define the alias as the key and the service ID as the value.
+--
+-- eg 1. the service ID ["F000"] = "F000" enables Envoy to forward requests with the subdomain "F000.path.grove.city" to PATH with the service ID "F000".
+-- eg 2. the alias ["pocket"] = "F000" enables Envoy to forward requests with the subdomain "pocket.path.grove.city" to PATH with the service ID "F000".
+--
+-- To utilize PATH's Quality of Service (QoS) features, the service ID must match the value in PATH's `qos` module.
+-- TODO_IMPROVE(@commoddity): Add link to the file & line in the QoS module once 'no-op' QoS feature is completed.
```
Olshansk, line 12: Can we do this in this commit?

envoy/scripts/copy_envoy_config.sh
```diff
@@ -6,29 +6,101 @@
 # Get the directory of the script
```
Olshansk, line 6: Just as a heads up - I'm skipping this in leu of time but will edit when going through the E2E flow again.

Makefile
commoddity: Done; updated the test data to be consistent with the Gateway Endpoints example file in PADS.

envoy/allowed-services.template.lua
commoddity: Good catch. Done now; just merged this file in in the last PR so now it's pointing to the right place.

envoy/scripts/copy_envoy_config.sh
commoddity: > Just as a heads up - I'm skipping this in leu of time but will edit when going through the E2E flow again.

SGTM but just FYI this file it not currently involved in the E2E at all. 

I started playing around with how we could implement a FULL E2E suite utilizing Tilt to spin up the full stack including Envoy but it's non-trivial and will have to wait until the new year. 

But just FYI I have been thinking about it and tentatively started prototyping.

Olshansk (approved):
@commoddity I updated the docs and made sure we can render mermaid diagrams, etc (it wasn't working before).

Let's merge it in!

PR #110, opened 2024-12-18
fix: small changes to E2E tests to use header for passing service ID

commoddity:
Updates to remove deprecated usage of `subdomain` for passing service ID in favour of passing it in the config.

Also cleans up the example config files that are used in the E2E tests to remove references to `alias`.


e2e/morse_relay_test.go
```diff
@@ -24,26 +25,28 @@ func Test_MorseRelay(t *testing.T) {
...
 		{
-			name:         "should successfully relay eth_chainId for eth (F00C)",
-			reqMethod:    http.MethodPost,
-			serviceAlias: "F00C",
-			relayID:      "1201",
-			body:         `{"jsonrpc": "2.0", "id": "1201", "method": "eth_chainId"}`,
+			name:      "should successfully relay eth_chainId for eth (F00C)",
+			reqMethod: http.MethodPost,
+			serviceID: "F00C",
+			reqPath:   "/v1/abcdef12",
```
adshmh, line 39: This should not be made part of the test case, as it is the same for all test cases.

commoddity: Yes but we may want the functionality to run the test for other services, which is why I added it to the test variables.

e2e/shannon_relay_test.go
```diff
@@ -37,27 +39,28 @@ func Test_ShannonRelay(t *testing.T) {
 			// single endpoint, maintained by Grove.
 			name:      "should successfully relay eth_blockNumber for anvil",
 			reqMethod: http.MethodPost,
+			serviceID: "anvil",
+			reqPath:   "/v1/abcdef12",
```
adshmh, line 43: Ditto here: `reqPath` is the same for all the test cases. If at some point that changes, we can update the tests accordingly.

e2e/shannon_relay_test.go
```diff
@@ -37,27 +39,28 @@ func Test_ShannonRelay(t *testing.T) {
...
 			relayID:   "1002",
 			body:      `{"jsonrpc": "2.0", "id": "1002", "method": "eth_chainId"}`,
 		},
 		// TODO_UPNEXT(@adshmh): add more test cases with valid and invalid jsonrpc request payloads.
 	}
 
-	reqPath := "/v1/abcdef12"
-	serviceID := "anvil"
```
adshmh, line 53: I don't think this should be moved to individual tests: `anvil` is going to be used for Shannon E2E tests for now.

adshmh (changes requested):
Thank you for the PR. Just a few minor blockers to address.

e2e/shannon_relay_test.go
commoddity: > Ditto here: `reqPath` is the same for all the test cases. If at some point that changes, we can update the tests accordingly.

I take your point but curious what the harm is in having them as test variables vs singlularly defined vars? My preference is generally to have any variables used in the test defined in the test cases for both consistency and the possibility of using other values down the line.

e2e/morse_relay_test.go
adshmh: PATH does not differentiate between requests based on this part of the URL at this point. E2E tests need to make this clear by explicitly declaring the value as a single variable/constant used by all the tests.

commoddity: OK for `reqPath` you're right and that makes sense. But for the service ID IMO it's worth keeping as a test far.

e2e/shannon_relay_test.go
adshmh: request path is not really `variable` here: all the tests use the same value. Declaring it as such makes it a lot easier to read the E2E tests by anyone who is not familiar with how PATH treats different parts of the request URL.

commoddity: > request path is not really `variable` here: all the tests use the same value. Declaring it as such makes it a lot easier to read the E2E tests by anyone who is not familiar with how PATH treats different parts of the request URL.

Got it; I do agree with this re: the request path.

adshmh (approved):

PR #112, opened 2024-12-19
[WIP] Integrate noop service into PATH

Olshansk:
## Issue

- #{ISSUE_NUMBER}

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [ ] **All Tests**: `make test_all`
- [ ] **Documentation**: `make docusaurus_start`; only needed if you make doc changes

## Sanity Checklist

- [ ] I have tested my changes using the available tooling
- [ ] I have commented my code
- [ ] I have performed a self-review of my own code; both comments & source code
- [ ] I create and reference any new tickets, if applicable
- [ ] I have left TODOs throughout the codebase, if applicable

### DIFF
--- a/cmd/qos.go
+++ b/cmd/qos.go
@@ -9,6 +9,7 @@ import (
 	"github.com/buildwithgrove/path/gateway"
 	"github.com/buildwithgrove/path/protocol"
 	"github.com/buildwithgrove/path/qos/evm"
+	"github.com/buildwithgrove/path/qos/noop"
 )
 
 // TODO_UPNEXT(@adshmh): enable Solana QoS instance through the following steps:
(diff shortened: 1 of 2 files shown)


cmd/qos.go
```diff
@@ -74,6 +75,11 @@ func getServiceQoSInstances(
 		case config.ServiceIDPOKT:
 			// TODO_TECHDEBT: add pokt qos service here
 
+		case config.ServiceIDNoop:
+			if _, ok := gatewayServiceIDsIdx[serviceID]; ok {
+				gatewayQoSService[serviceID] = noop.NoOpQoS{}
+			}
+
 		default:
 			return nil, nil, fmt.Errorf("error building QoS instances: service ID %q not supported by PATH", serviceID)
```
RawthiL, line 84: Why isn't the default QoS an instance of `noop.NoOpQoS{}` instead returning an error?
I would prefer a big warning and allow the user to proceed without any QoS module.

If I understand correctly the map of service IDs to QoS modules is hardcoded. If I stake a new service that is not in the list of the NoOpQoS, I would need to manually add my service to the NoOpQoS list and recompile PATH.

PR #106, opened 2024-12-17
[QoS] Introduce v0 of Noop QoS module

adshmh:

PR #113, opened 2024-12-19
[QoS] feat: expand support for NoOp functionality

commoddity:
# tl;dr

This PR makes PATH default to a NoOp QoS for services without a QoS implementation instead of rejecting requests, streamlining configuration and simplifying PATH's service config to focus on QoS mappings only.

This PR also removes the requirement for defining services in the PATH config YAML, as the responsibility for "allowed services" now resides with Envoy Proxy (this change is handled in PR #102 ).

_Example Test_
Screenshot 2024-12-19 at 11 09 20

## Summary

This PR expands upon the draft functionality introduced by @Olshansk in PR https://github.com/buildwithgrove/path/pull/112, which itself builds upon @adshmh 's work in https://github.com/buildwithgrove/path/pull/106.

It achieves several primary goals (see reasoning for these changes in the next section):

1. Removes the requirement for the `services` field in the config YAML structure. 
2. Allows the `Parser` struct to default to the `NoOp` QoS implementation if the request's service ID does not have a corresponding QoS implementation.
3. Renames `config/service_alias.go` to `config/service_qos.go`
4. As a direct requirement of [1], it updates the `cmd/qos.go` file's initialization logic to create the `gatewayQoSService` map from the values in `config/service_qos.go` instead of from the services specified in the config YAML.
5. Improves commented documentation related to this change, including introducing the new "authoritative service ID" terminology.

## Reasoning

The reasoning for this change is as follows:

  1. PATH's current behaviour is to force operators to define which services their PATH instance supports in the config YAML file. This includes defining a string `alias` for service IDs (eg. `eth` aliases the authoritative service ID `F00C`).
  2. In discussions with @adshmh in Discord and Github it was decided that the responsibility for both defining service aliases and determining which service IDs PATH supports should be moved to the Envoy Proxy level.
  3. This means PATH only receives the "authoritative service ID" (eg. `F00C`, `anvil`, etc.), never an alias and it only ever receives it as the `target-service-id` header, never as the subdomain. 
  4. The change to handle this at the Envoy Level is implemented in PR#102: https://github.com/buildwithgrove/path/pull/102. This PR contains work parallel and related to this PR.
  5. As described in [1], PATH's current behaviour is to reject requests for services not configured in both the config YAML and the `config/service_qos.go` (previously `config/service_alias.go`) file. 
  6. With the introduction of the `NoOp QoS` functionality, this has has changed to instead use a QoS service that selects a random endpoint and performs no QoS checks, rather than outright rejecting the request.
  7. Because of [2] and [4], the source of truth for "allowed services" will soon reside at the Envoy level. This means that the `config/service_qos.go` should be concerned only with which services correspond to which QoS implementations (eg. `F00C -> EVM QoS`, `proto-anvil -> NoOp QoS`, etc.) and not with whether the service is "allowed" by PATH.
  8. This also means that we no longer need to configure which services are allowed by the PATH instance in the config YAML as all services are "allowed" at the PATH level; forcing operators to configure services at both the Envoy and PATH level is confusing and redundant and should be removed. _(Note that the `hydrator_config` that determines which services the hydrator runs for remains in place and may still be configured by the operator; only the `services` YAML field is removed.)_

## Origin

Screenshot 2024-12-19 at 10 16 43
https://discord.com/channels/824324475256438814/1263851886486880356/1318695800754405466

Screenshot 2024-12-19 at 10 17 31
https://github.com/buildwithgrove/path/pull/106#pullrequestreview-2511292581

## Type of change

- [x] Code health or cleanup

## Testing

- [x] **All Tests**: `make test_all`

## Sanity Checklist

- [x] I have tested my changes using the available tooling
- [x] I have commented my code
- [x] I have performed a self-review of my own code; both comments & source code
- [x] I create and reference any new tickets, if applicable

### DIFF
--- a/cmd/qos.go
+++ b/cmd/qos.go
@@ -17,7 +17,7 @@ import (
 // 2. Build a Solana QoS instance using any required configuration options.
 // 3. Pass the Solana QoS instance to the endpoint hydrator, if enabled.
 // 4. Pass the Solana QoS instance to the gateway.
-//
+
 // getServiceQoSInstances returns all QoS instances
 // to be used by the Gateway and EndpointHydrator, respectively.
 // This is done to ensure the same QoS instance is used in both
@@ -31,24 +31,20 @@ func getServiceQoSInstances(
 	map[protocol.ServiceID]gateway.QoSEndpointCheckGenerator,
 	error,
 ) {
-	// Build a map of services configured for the hydrator
-	// and the gateway to allow easy lookup.
+	// Build a map of services configured for the hydrator to allow easy lookup.
 	hydratorServiceIDsIdx := buildServiceIDsIdx(gatewayConfig.HydratorConfig.ServiceIDs)
-	gatewayServiceIDsIdx := buildServiceIDsIdx(gatewayConfig.GetEnabledServiceIDs())
 
-	// TODO_TECHDEBT: refactor this function to remove
-	// the need for manually adding entries for every new QoS implmenetation.
+	// TODO_TECHDEBT(@adshmh): refactor this function to remove the
+	// need to manually add entries for every new QoS implementation.
 	gatewayQoSService := make(map[protocol.ServiceID]gateway.QoSService)
 	hydratorQoSGenerators := make(map[protocol.ServiceID]gateway.QoSEndpointCheckGenerator)
 
-	// TODO_FUTURE: support serviceQoS-specific configuration.
-	allServiceIDs := append(gatewayConfig.GetEnabledServiceIDs(), gatewayConfig.HydratorConfig.ServiceIDs...)
+	// Initialize NoOp QoS service to handle services without a corresponding QoS implementation.
+	gatewayQoSService[protocol.ServiceIDNoOp] = noop.NoOpQoS{}
 
-	for _, serviceID := range allServiceIDs {
-
-		// Get the QoS type for the service ID.
-		serviceQoSType := config.ServiceQoSTypes[serviceID]
-
+	// Initialize QoS services for all service IDs with a corresponding QoS
+	// implementation, as defined in the `config/service_qos.go` file.
+	for serviceID, serviceQoSType := range config.ServiceQoSTypes {
 		switch serviceQoSType {
 
 		case config.ServiceIDEVM:
@@ -59,9 +55,10 @@ func getServiceQoSInstances(
 				},
 				Logger: logger,
 			}
-			if _, ok := gatewayServiceIDsIdx[serviceID]; ok {
-				gatewayQoSService[serviceID] = evm.NewServiceQoS(evmEndpointStore, logger)
-			}
+			gatewayQoSService[serviceID] = evm.NewServiceQoS(evmEndpointStore, logger)
+
+			// If the service is configured for the hydrator in the configuration YAML file,
+			// add its authoritative service ID to the hydrator's QoS generators map.
 			if _, ok := hydratorServiceIDsIdx[serviceID]; ok {
 				hydratorQoSGenerators[serviceID] = evmEndpointStore
 			}
@@ -75,12 +72,7 @@ func getServiceQoSInstances(
 		case config.ServiceIDPOKT:
 			// TODO_TECHDEBT: add pokt qos service here
 
-		case config.ServiceIDNoop:
-			if _, ok := gatewayServiceIDsIdx[serviceID]; ok {
-				gatewayQoSService[serviceID] = noop.NoOpQoS{}
-			}
-
-		default:
+		default: // this should never happen
 			return nil, nil, fmt.Errorf("error building QoS instances: service ID %q not supported by PATH", serviceID)
 		}
 	}
--- a/cmd/qos.go
+++ b/cmd/qos.go
@@ -9,7 +9,6 @@ import (
 	"github.com/buildwithgrove/path/gateway"
 	"github.com/buildwithgrove/path/protocol"
 	"github.com/buildwithgrove/path/qos/evm"
-	"github.com/buildwithgrove/path/qos/noop"
 )
 
 // TODO_UPNEXT(@adshmh): enable Solana QoS instance through the following steps:
@@ -38,9 +37,6 @@ func getServiceQoSInstances(
 	// need to manually add entries for every new QoS implementation.
 	gatewayQoSService := make(map[protocol.ServiceID]gateway.QoSService)
 	hydratorQoSGenerators := make(map[protocol.ServiceID]gateway.QoSEndpointCheckGenerator)
-
-	// Initialize NoOp QoS service to handle services without a corresponding QoS implementation.
-	gatewayQoSService[protocol.ServiceIDNoOp] = noop.NoOpQoS{}
 
 	// Initialize QoS services for all service IDs with a corresponding QoS
 	// implementation, as defined in the `config/service_qos.go` file.
--- a/cmd/qos.go
+++ b/cmd/qos.go
@@ -9,6 +9,7 @@ import (
 	"github.com/buildwithgrove/path/gateway"
 	"github.com/buildwithgrove/path/protocol"
 	"github.com/buildwithgrove/path/qos/evm"
+	"github.com/buildwithgrove/path/qos/noop"
 )
 
 // TODO_UPNEXT(@adshmh): enable Solana QoS instance through the following steps:
@@ -73,6 +74,11 @@ func getServiceQoSInstances(
 
 		case config.ServiceIDPOKT:
 			// TODO_TECHDEBT: add pokt qos service here
+
+		case config.ServiceIDNoop:
+			if _, ok := gatewayServiceIDsIdx[serviceID]; ok {
+				gatewayQoSService[serviceID] = noop.NoOpQoS{}
+			}
 
 		default:
 			return nil, nil, fmt.Errorf("error building QoS instances: service ID %q not supported by PATH", serviceID)
--- a/config/config.go
+++ b/config/config.go
@@ -2,7 +2,6 @@ package config
 
 import (
 	"errors"
-	"fmt"
 	"os"
 	"time"
 
@@ -10,8 +9,6 @@ import (
 
 	"github.com/buildwithgrove/path/config/morse"
 	"github.com/buildwithgrove/path/config/shannon"
-	"github.com/buildwithgrove/path/config/utils"
-	"github.com/buildwithgrove/path/protocol"
 )
 
 /* ---------------------------------  Gateway Config Struct -------------------------------- */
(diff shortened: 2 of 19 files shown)


qos/noop/context.go
```diff
@@ -67,7 +67,7 @@ func (rc *requestContext) GetHTTPResponse() gateway.HTTPResponse {
 
 	return &HTTPResponse{
 		httpStatusCode: http.StatusOK,
-		payload:        rc.receivedResponses[len(rc.receivedResponses)].ResponseBytes,
+		payload:        rc.receivedResponses[len(rc.receivedResponses)-1].ResponseBytes,
```
adshmh, line 70: Thanks for fixing this!

protocol/service.go
```diff
@@ -6,3 +6,7 @@ package protocol
 // B. The `qos` package handles offchain specs & details.
 // C. This allows the `gateway` package to map multiple Service IDs to a single qos implementation, e.g. all EVM blockchain services can be handled by `qos/evm`.
 type ServiceID string
+
+// ServiceIDNoOp is the Service ID for the NoOp QoS service.
+// This is used when a service ID does not yet have a QoS implementation.
+const ServiceIDNoOp ServiceID = "noop"
```
adshmh, line 12: Strong opinion: this does not belong here. `protocol` should not in any way be aware of or care about offchain QoS, and PATH has been designed to maintain this restriction.

Mapping of onchain service IDs, AKA authoritative service IDs, to specific QoS implementations is the responsibility of the `cmd` and/or `config` packages (depending on the desired level of flexibility).

If some sort of validation/authority from the `protocol` package is desired to e.g. validate user config/input, we can extend `gateway/qos.go` file to add a method under `QoSService`, e.g.:
```go
type  QoSService interface {
	QoSContextBuilder
        func ValidServiceIDs() ([]protocol.ServiceID, error)
}
```

cmd/qos.go
```diff
@@ -31,24 +31,20 @@ func getServiceQoSInstances(
...
 
-	// TODO_FUTURE: support serviceQoS-specific configuration.
-	allServiceIDs := append(gatewayConfig.GetEnabledServiceIDs(), gatewayConfig.HydratorConfig.ServiceIDs...)
-
-	for _, serviceID := range allServiceIDs {
-
-		// Get the QoS type for the service ID.
-		serviceQoSType := config.ServiceQoSTypes[serviceID]
+	// Initialize NoOp QoS service to handle services without a corresponding QoS implementation.
+	gatewayQoSService[protocol.ServiceIDNoOp] = noop.NoOpQoS{}
```
adshmh, line 43: Please see my comment re leaking `ServiceIDNoOp` into the `protocol` package.

Could you please confirm if this is the desired behavior: 

1. OPTIONAL VALIDATION: query the protocol on valid/authoritative service IDs, and reject the request if it the user-specified service ID is not recognized by the protocol.
2. If the user-specified service ID is mapped to a specific QoS instance, use that instance.
3. Otherwise, use the (singleton) NoOp QoS instance.

If the above is correct: no initialization is needed here: `request` package can simply return a/the NoOp QoS instance when parsing a user request.

protocol/service.go
commoddity: No problem; I have absolutely zero attachment to this variable living here.

The main point is just for us to have a consistent key value to be able to retrieve the NoOp QoS service from the `gatewayQoSService` map. I'll provide more detail in answer to your next comment.

cmd/qos.go
```diff
@@ -59,9 +55,10 @@ func getServiceQoSInstances(
 				},
 				Logger: logger,
 			}
-			if _, ok := gatewayServiceIDsIdx[serviceID]; ok {
-				gatewayQoSService[serviceID] = evm.NewServiceQoS(evmEndpointStore, logger)
-			}
+			gatewayQoSService[serviceID] = evm.NewServiceQoS(evmEndpointStore, logger)
```
adshmh, line 58: This PR is dealing with multiple aspects, and we need to focus on a single one in each PR to make it easier for any future readers to follow;
- Initialization of QoS instances for their corresponding recognized service IDs: this PR should not change anything here.
- Desired behavior re NoOp QoS, i.e. service IDs that are valid from the perspective of protocol, but have not been configured, specified, embedded, etc.: this seems to be the main target of the PR, let's focus on it exclusively.

cmd/qos.go
commoddity: > Please see my comment re leaking `ServiceIDNoOp` into the `protocol` package.
Agreed; we can move the const anywhere (or decide on a different way to retrieve `noop.NoOpQoS{}` from the map.
> 1. OPTIONAL VALIDATION: query the protocol on valid/authoritative service IDs, and reject the request if it the user-specified service ID is not recognized by the protocol.
This could be done in the future; I would argue it's outside the scope of this PR but not opposed to doing it later.
> 2. If the user-specified service ID is mapped to a specific QoS instance, use that instance.
Correct.
> 3. Otherwise, use the (singleton) NoOp QoS instance.
Also correct.
> If the above is correct: no initialization is needed here: `request` package can simply return a/the NoOp QoS instance when parsing a user request.
But how do we retrieve `noop.NoOpQoS{}` from the `gatewayQoSService` map? Being able to retrieve the NoOp service QoS instance was the intention behind the `ServiceIDNoOp` const (which again, can be moved if desired).

config/config.go
```diff
@@ -25,13 +22,9 @@ type (
 		MorseConfig   *morse.MorseGatewayConfig     `yaml:"morse_config"`
 		ShannonConfig *shannon.ShannonGatewayConfig `yaml:"shannon_config"`
 
-		Services        map[protocol.ServiceID]ServiceConfig `yaml:"services"`
```
adshmh, line 28: Not opposed to removing `Services`: letting `EnvoyProxy` handle what service(s) the users can request seems like a cleaner solution.

Do you think a configurable mapping of Authoritative service IDs to QoS implementations makes sense?
e.g.:
```yaml
qos_service_id_mappings:
  "evm":
    - "F001"
    - "F00C"
  "solana":
    - "S001"
    - "SOL"
```

config/config.go
```diff
@@ -75,43 +64,8 @@ func (c GatewayConfig) GetRouterConfig() RouterConfig {
...
-	var enabledServices []protocol.ServiceID
-	for serviceID := range c.Services {
-		enabledServices = append(enabledServices, serviceID)
-	}
-	return enabledServices
-}
-
 /* --------------------------------- Gateway Config Hydration Helpers -------------------------------- */
 
-func (c *GatewayConfig) hydrateServiceAliases() error {
```
adshmh, line 100: Great idea removing service aliases as a concern of PATH.

cmd/qos.go
commoddity: > This PR is dealing with multiple aspects, and we need to focus on a single one in each PR to make it easier for any future readers to follow;
> 
> * Initialization of QoS instances for their corresponding recognized service IDs: this PR should not change anything here.
> * Desired behavior re NoOp QoS, i.e. service IDs that are valid from the perspective of protocol, but have not been configured, specified, embedded, etc.: this seems to be the main target of the PR, let's focus on it exclusively.

IMO these two issues are strongly related: if we remove the limitation that a service must be specified in the config YAML **and** `service_alias.go` (now `service_qos.go`) then there is no logical reason to force an operator to configure the `services` field in the YAML when we already hardcode the mapping of `service ID -> QoS` and can derive the `gatewayQoSService` map from there.

Forcing an operator to specify this yaml field adds an unnecessary layer of configuration and mental overhead for no clear benefit. 

If the argument is just that it should be split into two separate PRs I can understand that but this work is already done and functioning in this PR; I can split it if you like but currently this functionality is blocking multiple potential gateway operators from using PATH so I'm a bit hesitant to split already completed work into multiple parts.

adshmh (changes requested):
@commoddity Great PR, thank you!
I have stopped after a first high-level review iteration, because I think some of the comments need further discussion before proceeding with the implementation details.

config/config.go
commoddity: To be honest I think it would be better to have one authoritative source of truth for which service uses which QoS. Adding this seems like additional configuration for an unclear purpose.

Is there a use case where an operator would want requests for `Ethereum` to be handled by the QoS service for `Solana`? IMO it's pretty clear cut which QoS is for which service and they are not interoperable.

Even if there does arise a use case wherein interoperable QoS is desired and useful I think we are a long way from there and so I'd recommend not adding this to scope of this PR at this stage.

cmd/qos.go
adshmh: No need to retrieve a `noop.NoOpQoS{}` here:
To fulfill `3` from above, the `request.Parser` can have a default behavior or returning a `noop.NoOpQoS{}` if it does not recognize the user-requested service ID.

adshmh: Please note that returning a QoS instance to match a user HTTP request, is solely the responsibility of the `request.Parser` struct:
image

commoddity: > To fulfill 3 from above, the request.Parser can have a default behavior or returning a noop.NoOpQoS{} if it does not recognize the user-requested service ID.

Sure makes sense. Are there no performance implications with doing so? If not I will update the PR to do this.

> Please note that returning a QoS instance to match a user HTTP request, is solely the responsibility of the `request.Parser` struct:
> 
> image
> 
> 

No argument here; that's not contradicted by what the PR is already doing.

commoddity: @adshmh Please note I have removed the `ServiceIDNoOp` const and updated the `Parser.GetQoSService` method:
```go
// GetQoSService returns the QoS service implementation for the given request, as well as the authoritative service ID.
// If the service ID does not have a corresponding QoS implementation, the NoOp QoS service is returned.
func (p *Parser) GetQoSService(ctx context.Context, req *http.Request) (protocol.ServiceID, gateway.QoSService, error) {
	// Get the authoritative service ID from the request's header.
	serviceID, err := p.getServiceID(req)
	if err != nil {
		return "", nil, err
	}

	// Get the QoS service implementation for the request's service ID if it exists.
	if qosService, ok := p.qosServices[serviceID]; ok {
		return serviceID, qosService, nil
	}

	// If the service ID is not supported, use the NoOp QoS service, which will select a random endpoint.
	// The NoOp QoS service is initialized in the `cmd/qos.go` file on startup of PATH.
	return serviceID, &noop.NoOpQoS{}, nil
}
```

See commit: https://github.com/buildwithgrove/path/pull/113/commits/36b3e69e555187fdf1f0e299e73adac0a885da95

adshmh: SGTM, please just add a `TODO_TECHDEBT(@dashmh)` to validate the service ID using a protocol instance.

config/config.go
adshmh: OK, no strong opinions on this from me. If this turns out to be a high-in-demand feature, we will revisit it.

adshmh (approved):
Pre-approving: please add the requested TODO item before merging.

Olshansk (approved):
Great work! LGTM :shipit:

PR #122, opened 2025-01-02
fix: small tweaks to improve localnet dev experience

commoddity:
## Summary

A couple small tweaks to the localnet configuration for improved developer experience.

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -109,8 +109,8 @@ if MODE == "path_with_auth":
     # ---------------------------------------------------------------------------- #
     #                             Envoy Auth Resources                             #
     # ---------------------------------------------------------------------------- #
-    # 1. External Auth Server                                                      #
-    # 2. Envoy Proxy                                                               #
+    # 1. Envoy Proxy                                                               #
+    # 2. External Auth Server                                                      #
     # 3. Path Auth Data Server (PADS)                                              #
     # 4. Rate Limiter                                                              #
     # 5. Redis                                                                     #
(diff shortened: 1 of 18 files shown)


envoy/envoy.template.yaml
```diff
@@ -113,42 +113,43 @@ static_resources:
                   - name: envoy.filters.http.lua
                     typed_config:
                       "@type": type.googleapis.com/envoy.extensions.filters.http.lua.v3.Lua
-                      inline_code: |
```
commoddity, line 116: `inline_code` was a deprecated field.

Ref:
https://www.envoyproxy.io/docs/envoy/latest/api-v3/extensions/filters/http/lua/v3/lua.proto#extensions-filters-http-lua-v3-lua

Tiltfile
```diff
@@ -103,7 +103,6 @@ else:
             "ratelimit",
             "redis",
         ],
-        port_forwards=["3000:3000"],
```
commoddity, line 106: PATH running with Envoy should not expose port 300 as all traffic should go through Envoy on port 3001.

Tiltfile
```diff
@@ -139,7 +138,18 @@ if MODE == "path_with_auth":
...
 
-    # 1. Build the External Auth Server image from envoy/auth_server/Dockerfile
+    # 1. Load the Kubernetes YAML for the envoy-proxy service
+    k8s_yaml("./local/kubernetes/envoy-proxy.yaml")
+    k8s_resource(
+        "envoy-proxy",
+        labels=["envoy_auth"],
+        # By default the Envoy Proxy container will bind to 127.0.0.1:3001.
+        # If you wish Envoy Proxy to be acessible outside the host machine,
+        # you must specify `port_forwards=["0.0.0.0:3001:3001"]`.
```
Olshansk, line 148: Any reason we don't want to just make this the default for local development?

local/kubernetes/path-values.yaml
```diff
@@ -28,15 +28,21 @@ path:
       service:
         type: ClusterIP
 
+  # The PATH service uses the `/healthz` endpoint to check if the service is ready,
+  # which uses the response from the QoS service to determine whether to send a 
+  # 503 (Service Unavailable) or 200 (OK) status code.
+  #
+  # When using PATH with large numbers of services enabled, the QoS service
+  # may take a long time to return a healthy response. Therefor, it is expected
```
Olshansk, line 36:
```suggestion
  # may take a long time to return a healthy response. Therefore, it is expected
```

Olshansk (approved):

Tiltfile
commoddity: No real reason; I will do that.

Issue #124, opened 2025-01-03
[Websocket] PATH Only (no onchain) Websocket Prototype


Olshansk:
## Objective

Build the foundation for WebSockets in PATH

## Origin Document

Lots of customers have been waiting on this :)

[Design Document Started in Notion](https://www.notion.so/buildwithgrove/PATH-Websockets-Design-174a36edfff680358b5fc1c87725b5f0#174a36edfff680148b7dddf5b43f752e)

## Goals

- Set the foundation for WebSockets in PATH
- Unblock new customers and business deals
- Identify missing gaps in our knowledge and understanding of how to implement 

## Deliverables

- [ ] 1. Implement a mega throwaway PR that prototypes websockets from start to end w/ the goal of functionality but code does not have to be clean.
- [ ] 2.1 Collect learnings from (1) and get some feedback from the team w.r.t the approach, design, open questions and next steps
- [ ] 2.2 Compare and align our WebSocket interface with others (e.g. alchemy) for consistency
- [ ] 3. Break apart (1) into smaller PRs that'll be merged into main after (2) is complete
- [ ] 4. Prepare a design doc that'll be merged in path.grove.city 
- [ ] 5. Provide tools (e.g. Makefile targets) to make this easy to test & use
- [ ] 6. Add E2E websocket tests in CI
- [ ] 7. Align w/ @red-0ne on next steps to implement websockets onchain

## Non-goals / Non-deliverables

- Making websockets prod ready
- Deploying websockets to prod
- Implementing onchain websockets
- WebRTC compatibility
- SSE (Server Side Events) compatability

---

**Creator**: @Olshansk 
**Co-Owners**: @commoddity @red-0ne

PR #135, opened 2025-01-15
[WebSockets] Websocket override URLs

commoddity:
## Summary

This PR lays the groundwork for enabling WebSockets in PATH.

**🚨 NOTE: The implementation in this PR DOES NOT enable onchain WebSockets as Shannon does not yet support this.**

### 1. WebSocket Endpoint URLs

The implementation in this PR uses a new field in the config YAML:
`shannon_config.websocket_endpoint_urls`

This field is passed to the `Gateway` struct and allows PATH to open a WebSocket connection to this user-specified websocket-enabled URL. One endpoint URL may be specified per `service ID`.

eg.
```yaml
shannon_config:
  ...
  websocket_endpoint_urls: 
    "anvil": wss://shannon-testnet-grove-websocket.beta.poktroll.com
```

Due to the fact that this connection is direct to a user-specified Endpoint URL, this connection bypasses almost everything in the current service request lifecycle:
- Request parsing
- Request contexts (QoS, gateway and protocol contexts)
- Endpoint selection

_Example Request_
```bash
wscat -c ws://localhost:3069/v1/ -H "target-service-id: anvil"
```

When a WebSocket request is sent, PATH's `gateway` package opens a `bridge` (websocket connection from `Client <--> PATH <--> User-specified off-chain endpoint URL`) which allows for a persistent WebSocket connection for handling two-way communication between the Endpoint and client.

**As indicated in the comments in the code, this is a TEMPORARY solution to enable PATH to allow WebSockets connections and will be heavily modified once Shannon supports WebSocket connections.**

### 2. `websockets` Package

This PR also introduces a `websockets` package, which contains two key structs (and unit tests):

- `bridge`

As mentioned above, the `bridge` struct opens two connections, to enable communication as follows:
`Client <--> PATH <--> User-specified off-chain endpoint URL`

https://github.com/buildwithgrove/path/blob/websocket-passthrough/websockets/bridge.go

- `connection` 

The `connection` struct abstracts the logic of reading & writing to/from a websocket connection so that it can be reused for both Client and Endpoint connections in the `bridge`.

https://github.com/buildwithgrove/path/blob/websocket-passthrough/websockets/connection.go

### 3. Protocol WebSocket Methods

This PR also introduces methods to the **Shannon** protocol **ONLY**. (Morse does and will not support WebSockets.)

These methods are present on the `gateway.Protocol` interface, as well as on the `Shannon` protocol struct that satisfies this interface.

_Note that the Morse protocol contains the methods for the purposes of satisfying the interface ONLY; they will always return an error if called by a PATH instance on the Morse protocol._

**These methods are currently unused as of this PR but should get us **MOST** of the way to a functioning onchain WebSockets implementation once the Shannon protocol and SDK support WebSocket connections.**

### 4. Future Work

All TODOs related to future work on the WebSocket implementation are marked with:
`TODO_FUTURE(@commoddity)[WebSockets]`

## Issue

- [PATH Only (no onchain) Websocket Prototype](https://github.com/buildwithgrove/path/issues/124)

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Testing

- [x] **All Tests**: `make test_all`
- [ ] **Documentation**: `make docusaurus_start`; only needed if you make doc changes

## Sanity Checklist

- [x] I have tested my changes using the available tooling
- [x] I have commented my code
- [x] I have performed a self-review of my own code; both comments & source code
- [x] I create and reference any new tickets, if applicable
- [x] I have left TODOs throughout the codebase, if applicable


Olshansk (changes requested):
@commoddity Just a heads up that I'm still in the progress of reviewing this but am also pushing some (breaking) changes. 

No need to tackle it yet - I'll notify you when I do. 

My ask in the meantime is that you start working on **simple** documentation for path.grove.city on how websockets work.

Transfer over parts of the notion doc you already had!

qos/noop/noop.go
```diff
@@ -36,6 +36,15 @@ func (_ NoOpQoS) ParseHTTPRequest(_ context.Context, httpRequest *http.Request)
 	}, true
 }
 
+// ParseWebsocketRequest builds a request context from the provided WebSocket request.
+// This method implements the gateway.QoSService interface.
+// TODO_HACK(@commoddity, WebSockets): Utilize this method once the Shannon protocol supports websocket connections.
```
Olshansk, line 41: Maybe we should jsut return an error for noop?

qos/evm/evm.go
```diff
@@ -56,6 +56,20 @@ func (qos *QoS) ParseHTTPRequest(_ context.Context, req *http.Request) (gateway.
 	}, true
 }
 
+// ParseWebsocketRequest builds a request context from the provided WebSocket request.
+// WebSocket connection requests do not have a body, so we don't need to parse it.
+//
+// This method implements the gateway.QoSService interface.
+// TODO_HACK(@commoddity, WebSockets): Utilize this method once the Shannon protocol supports websocket connections.
```
Olshansk, line 63: @commoddity Can you create a ticket to productionize websockets and then replace all `TODO_HACK(@commoddity, WebSockets)` with `TODO_HACK(@commoddity, #XXX)`?

Olshansk: Feel free to use a screenshot of this thread for reference + link to #135

protocol/shannon/context.go
```diff
@@ -98,6 +102,34 @@ func (rc *requestContext) HandleServiceRequest(payload protocol.Payload) (protoc
 	return relayResponse, nil
 }
 
+// HandleWebsocketRequest opens a persistent websocket connection to the selected endpoint.
+// Satisfies the gateway.ProtocolRequestContext interface.
+// TODO_HACK(@commoddity, WebSockets): Utilize this method once the Shannon protocol supports websocket connections.
+func (rc *requestContext) HandleWebsocketRequest(req *http.Request, w http.ResponseWriter, logger polylog.Logger) error {
```
Olshansk, line 108: NIT: Please make the `logger` the first field in the function.

protocol/shannon/context.go
```diff
@@ -98,6 +102,34 @@ func (rc *requestContext) HandleServiceRequest(payload protocol.Payload) (protoc
...
+	var upgrader = websocket.Upgrader{CheckOrigin: func(r *http.Request) bool { return true }}
+	clientConn, err := upgrader.Upgrade(w, req, nil)
+	if err != nil {
+		logger.Error().Err(err).Msg("Error upgrading websocket connection request")
+		return err
+	}
+
+	bridge, err := websockets.NewBridge(logger, selectedEndpointURL, clientConn)
+	if err != nil {
+		return err
```
Olshansk, line 123: Can you log errors here too

protocol/shannon/context.go
```diff
@@ -98,6 +102,34 @@ func (rc *requestContext) HandleServiceRequest(payload protocol.Payload) (protoc
...
+func (rc *requestContext) HandleWebsocketRequest(req *http.Request, w http.ResponseWriter, logger polylog.Logger) error {
+	var selectedEndpointURL string
+	if rc.selectedEndpoint != nil {
+		selectedEndpointURL = rc.selectedEndpoint.PublicURL()
+	}
+
+	var upgrader = websocket.Upgrader{CheckOrigin: func(r *http.Request) bool { return true }}
+	clientConn, err := upgrader.Upgrade(w, req, nil)
+	if err != nil {
+		logger.Error().Err(err).Msg("Error upgrading websocket connection request")
```
Olshansk, line 117: Do you mind adding a bit more details so tha it's easy for us to identify where the error happened when we're debugging? E.g. service_id, app_id, supplier_id?

Something for humans.

protocol/shannon/context.go
```diff
@@ -98,6 +102,34 @@ func (rc *requestContext) HandleServiceRequest(payload protocol.Payload) (protoc
...
+	}
+
+	bridge, err := websockets.NewBridge(logger, selectedEndpointURL, clientConn)
+	if err != nil {
+		return err
+	}
+
+	go bridge.Run()
+
+	logger.Info().Msg("websocket connection established")
```
Olshansk, line 128: Ditto regarding logger info.

qos/solana/solana.go
```diff
@@ -63,6 +63,20 @@ func (qos *QoS) ParseHTTPRequest(_ context.Context, req *http.Request) (gateway.
...
+// WebSocket connection requests do not have a body, so we don't need to parse it.
+//
+// This method implements the gateway.QoSService interface.
+// TODO_HACK(@commoddity, WebSockets): Utilize this method once the Shannon protocol supports websocket connections.
+func (qos *QoS) ParseWebsocketRequest(_ context.Context) (gateway.RequestQoSContext, bool) {
+	return &requestContext{
+		EndpointStore: qos.EndpointStore,
+		Logger:        qos.Logger,
+
+		isValid: true,
```
Olshansk, line 76:
```suggestion
		Logger:        qos.Logger,
		
		EndpointStore: qos.EndpointStore,
		
		isValid: true,
```

Olshansk (changes requested):
@commoddity Pushed a few more minor changes & some comments.

1. PTAL at the comments
2. Please start adding documentation (can be a separate PR)
3. Please double check I didn't break anything (I haven't verified)
4. Start thinking about how we can move this onchain (e.g. an event is a relay); I might ask you to implement the poktroll component

5. I'm going to dive into the meat of the PR tomorrow

config/examples/config.shannon_example.yaml
```diff
@@ -42,6 +42,13 @@ shannon_config:
       # README: the application private key MUST BE replaced with the correct application private key secret
       - 40af4e7e1b311c76a573610fe115cd2adf1eeade709cd77ca31ad4472509d388
 
+  # [OPTIONAL] If wishing to enable websockets, set the websocket_endpoint_url. This is a TEMPORARY workaround
```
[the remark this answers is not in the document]
commoddity: You can just point your `.config.yaml` file to the local schema that has all the latest changes, like so:
Screenshot 2025-01-23 at 13 01 43

qos/evm/evm.go
commoddity: Done.

qos/noop/noop.go
commoddity: I don't think we should. The reason being that there may be WebSocket enabled services without an existing QoS implementation and should still allow them to function in that case, albeit with random endpoint selection. Returning an error would effectively make them unusable until we build a QoS implementation for all of them.

commoddity (commented):
Implemented suggestions in review so far.

websockets/bridge.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/gorilla/websocket"
+	"github.com/pokt-network/poktroll/pkg/polylog"
+)
+
+// bridge routes data between an Endpoint and a Client. One bridge represents
+// a single WebSocket connection between a Client and a WebSocket Endpoint.
+//
+// Full data flow: Client <------> PATH <------> WebSocket Endpoint
+type bridge struct {
```
Olshansk, line 15: Do you mind adding a a few comments to `endpointConn`, `clientConn`, etc...

It's very clean, but I have to think a lot and read the code to understand where/how they're being used.

Why are msgCHan and stopChan shared between the two?

Etc...

websockets/bridge.go
```diff
@@ -0,0 +1,116 @@
...
+		endpointConn: endpointConnection,
+		clientConn:   clientConnection,
+		msgChan:      msgChan,
+		stopChan:     stopChan,
+
+		logger: logger,
+	}, nil
+}
+
+/* ---------- Public method - Run Bridge ---------- */
```
Olshansk, line 66: Should w create an interface and enforce adherence so it's self explanatory?

websockets/bridge.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/pokt-network/poktroll/pkg/polylog"
+)
+
+// bridge routes data between an Endpoint and a Client. One bridge represents
+// a single WebSocket connection between a Client and a WebSocket Endpoint.
+//
+// Full data flow: Client <------> PATH <------> WebSocket Endpoint
+type bridge struct {
+	logger       polylog.Logger
+	endpointConn *connection
```
Olshansk, line 18: Something similar to `Full data flow: Client <------> PATH <------> WebSocket Endpoint` for endpoint/client conn would be great.

websockets/bridge.go
```diff
@@ -0,0 +1,123 @@
...
+
+// Run starts the bridge and establishes a bidirectional communication
+// through PATH between the Client and the selected websocket endpoint.
+//
+// Full data flow: Client <------> PATH <------> WebSocket Endpoint
+func (b *bridge) Run() {
+	// Start goroutine to read messages from message channel
+	go b.messageLoop()
+
+	b.logger.Info().Msg("bridge operation started successfully")
```
Olshansk, line 76: Can you think of some identifiable metadata (for human readers) to distinguish this log message?

E.g. a service_id, an app_id, etc?

websockets/bridge.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// If close signal is received, stop the bridge and close both connections
```
Olshansk, line 78: Feel free to reward, but I'm thinking something like this.

```suggestion
	// Keep the bridge open until a stop signal is received (i.e. block until told otherwise)
```

websockets/bridge.go
```diff
@@ -0,0 +1,123 @@
...
+		case msg := <-b.msgChan:
+			switch msg.source {
+			// If the message is from the Client connection, send it to the Endpoint
+			case messageSourceClient:
+				b.handleClientMessage(msg)
+			// If the message is from the Endpoint, send it to the Client
+			case messageSourceEndpoint:
+				b.handleEndpointMessage(msg)
+			}
+		}
```
Olshansk, line 105:
```suggestion
		case msg := <-b.msgChan:
			switch msg.source {

			// If the message is from the Client connection, send it to the Endpoint
			case messageSourceClient:
				b.handleClientMessage(msg)

			// If the message is from the Endpoint, send it to the Client
			case messageSourceEndpoint:
				b.handleEndpointMessage(msg)
			}
		}
```

websockets/bridge.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+}
+
+// handleClientMessage processes a message from the Client and sends it to the Endpoint
+func (b *bridge) handleClientMessage(msg message) {
+	if err := b.endpointConn.WriteMessage(msg.messageType, msg.data); err != nil {
```
Olshansk, line 111: Wdyt of adding DEBUG level logs here and below.

At the very least, might be nice in the early days when testing/developing to see the bidrectional messages coming in.

websockets/connection.go
```diff
@@ -0,0 +1,181 @@
...
+		// stopChan is already closed, do nothing
+	default:
+		// stopChan is still open, send the error
+		c.stopChan <- fmt.Errorf("error reading from %s connection: %w", source, err)
+	}
+}
+
+// pingLoop sends keep-alive ping messages to the connection and handles pong messages
+func (c *connection) pingLoop() {
+	ticker := time.NewTicker(pingPeriod)
```
Olshansk: #PUC what this is for

websockets/connection.go
```diff
@@ -0,0 +1,181 @@
...
+		c.SetPongHandler(func(string) error {
+			if err := c.SetReadDeadline(time.Now().Add(pongWait)); err != nil {
+				c.logger.Error().Err(err).Msg("failed to set pong handler read deadline")
+			}
+
+			return nil
+		})
+	}
+
+	initPingLoop()
```
Olshansk, line 187: I don't fully understand why this is a separate function?  It's not even called in a go-routine.

websockets/connection.go
```diff
@@ -0,0 +1,181 @@
...
+
+	for {
+		select {
+		case <-c.stopChan:
+			return
+
+		case <-ticker.C:
+			if err := c.WriteControl(websocket.PingMessage, nil, time.Now().Add(writeWait)); err != nil {
+				c.logger.Error().Err(err).Msg("failed to send ping to connection")
+				c.stopChan <- fmt.Errorf("failed to send ping to connection: %w", err)
```
Olshansk, line 197: Consider update the logging so we have better visibility into the client / endpoint of this connection.

gateway/gateway.go
```diff
@@ -106,6 +127,80 @@ func (g Gateway) HandleHTTPServiceRequest(ctx context.Context, httpReq *http.Req
...
+// - Match HTTP request handling pattern
+// - Use HandleWebsocketRequest method defined on gateway.Protocol
+func (g Gateway) handleWebSocketRequest(ctx context.Context, httpReq *http.Request, w http.ResponseWriter) {
+	// Upgrade the HTTP request to a websocket connection.
+	// Do this first so that any errors that occur in the upgrade process can be sent
+	// to the websocket client as a close message, allowing easier debugging.
+	var upgrader = websocket.Upgrader{CheckOrigin: func(r *http.Request) bool { return true }}
+	clientConn, err := upgrader.Upgrade(w, httpReq, nil)
+	if err != nil {
+		g.Logger.Error().Msg("handleWebsocketRequest: error upgrading websocket connection request")
```
Olshansk, line 154: Will this have metadata as to which one?

Olshansk (changes requested):
@commoddity I pushed some NITs & edits, but left comments for the more functional changes.

We should be able to merge it right after!

gateway/gateway.go
commoddity: Yes, once the `protocol` package method are implemented the logs will contain information about which endpoint is selected to handle the connection.

For now it does not since each service only has one websocket URL, which is set in the config YAML.

commoddity:
> @commoddity I pushed some NITs & edits, but left comments for the more functional changes.
> 
> We should be able to merge it right after!

TY for the review. I have updated the comments in the `websockets` package with greater detail. Let me know if the changes work for you and if so I will merge.

websockets/connection.go
commoddity: That's a holdover from the previous implementation where it was called in multiple places (ie. on reconnect) but I've refactored the need for that out since we are offloading that responsibility to the client. So you are right it does not need to be a separate func; I have updated to remove the func var.

websockets/bridge.go
commoddity: Not sure if we need to since the `bridge` struct is only used in one possible way. My preference is not to define an interface here unless it's actually needed. That said, the comment is superfluous so I will remove it.

config/examples/config.shannon_example.yaml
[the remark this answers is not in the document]
Olshansk: Updaed some comments and pushed.

Olshansk (approved):

PR #148, opened 2025-01-23
[Workflows] Update GitHub actions from v3 to v4

Olshansk:
## Summary

Upgrade github ations from `v3` to `v4`.

### Origin Document

https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
Screenshot 2025-01-23 at 2 32 22 PM

### Example in action

Screenshot 2025-01-23 at 2 32 40 PM

https://github.com/buildwithgrove/path/actions/runs/12935368023/job/36078630477

### DIFF
--- a/.github/workflows/main-build.yml
+++ b/.github/workflows/main-build.yml
@@ -36,7 +36,7 @@ jobs:
             type=ref,event=branch,pattern=latest
 
       - name: Login to GitHub Container Registry
-        uses: docker/login-action@v4
+        uses: docker/login-action@v3
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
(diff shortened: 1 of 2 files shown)


.github/workflows/main-build.yml
```diff
@@ -36,7 +36,7 @@ jobs:
             type=ref,event=branch,pattern=latest
 
       - name: Login to GitHub Container Registry
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
```
okdas, line 39: v4 of `docker/login-action` doesn't exist: https://github.com/docker/login-action/releases

```suggestion
        uses: docker/login-action@v3
```

okdas (changes requested):

okdas (approved):
I think this should work now. Note: we don't use either of github actions mentioned in the screenshot, but I suspect they might be dependencies for other actions I bumped versions for. 

Let's see if everything works after merged to main.

fredteumer (approved):
Will mark this as i go through other repos to ensure this is bumped accordingly. thanks for flagging me

PR #149, opened 2025-01-24
[QoS][REST][CometBFT] Introduce CometQoS: Quality-of-Service for CometBFT & The Interchain

commoddity:
## Summary

Adds a CometBFT QoS implementation.

Spec Reference: https://docs.cometbft.com/v0.38/rpc

### QoS Checks

- `/health` - Simple node heartbeat. Returns empty response if node healthy or error response if node unhealthy.
- `/status` - Returns status including node info, pubkey, latest block hash, app hash, block height and time. **Only used to get latest block height.**

### Screenshots

The following screenshots show PATH functioning for `cometbft` on Shannon after `cometbft` service added with Supplier backend URL set to https://rpc.cosmos.directory/cosmoshub

**1. `/health` Request**
Screenshot 2025-01-24 at 19 26 33

**2. `/status` Request**
Screenshot 2025-01-24 at 19 26 50

**3. `header?height=24106587` Request**
Screenshot 2025-01-24 at 19 27 07

**4. `/header?height=24106587` Request (Envoy Proxy)**
Screenshot 2025-01-24 at 19 44 32

**5. Hydrator Enabled for `cometbft`** 
Screenshot 2025-01-24 at 19 24 47

### Changes:
- Adds `qos/cometbft` implementation. Built using same pattern as `qos/evm` package, then modified.
- Updates `cmd/qos.go` and `config/service_qos.go` files to include `cometbft`.

## Issue

- Description: Builds upon the work done in PR #145
- Issue: [#{ISSUE_NUMBER}](https://github.com/buildwithgrove/path/issues/87)
- Discussion: https://discord.com/channels/824324475256438814/1331712606330949764/1331744131097296993

## Type of change

Select one or more from the following:

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist

- [x] I have updated the GitHub Issue `assignees`, `reviewers`, `labels`, `project`, `iteration` and `milestone`
- [ ] For docs, I have run `make docusaurus_start`
- [ ] For code, I have run `make test_all`
- [ ] For configurations, I have update the documentation
- [ ] I added TODOs where applicable

PR #155, opened 2025-02-05
[Review] CometQoS Review

Olshansk:
Review of #149

Some notes:
- No real structural changes; just naming, nits, comments, etc...
- I tried to wrap my head around the whole thing and went too deep and out of scope
- Approach is good, but not amazing. We are setting the foundation for non-technical additions thought
- I understand the need for a refactor now
- We can likely use CometBFT as the "starting point" for the refactor, but it'll also require farther cleanup, even with this PR

Suggested next steps;
1. Review this PR
2. Merge it into #149 assuming it looks good
3. Add some makefile helpers in #149 specific to CometBFT
4. Merge the whole thing into main
5. Kickoff the refactor


commoddity (approved):
LGTM, all changes look positive to me. ✅

PR #163, opened 2025-02-13
[QoS] Add ability to set expiry for checks in EVM QoS

commoddity:
## Summary
Implements a refreshed EVM QoS checks interface with an expiry mechanism to better handle endpoint validations over time.

### Primary Changes:
- Introduce new per-check files (`check_blocknumber.go`, `check_chainid.go`, `check_empty.go`) that encapsulate EVM-specific validations.  
- Add an expiration policy to `chain_id` and `block_height` checks so they can re-run after a set interval.  

### Secondary changes:
- Remove the old `methods.go` in favor of specialized checks with `shouldRun()` and `isValid()` routines.  
- Refactor QoS data structures to include `syncAllowance` and reduce duplicated logic.
- Implements the below `TODO` in the `store.go/state.go` files:
```go
// TODO_MVP(@adshmh): rename the EndpointStoreConfig struct below and use it in the `State` struct.	
// The `EndpointStore` will only maintain data on the endpoints instead of how this data should be used	
// to validate endpoints.
```

## Type of change

- [ ] Bug fix
- [x] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## Sanity Checklist

- [ ] I have updated the GitHub Issue 'assignees', 'reviewers', 'labels', 'project', 'iteration' and 'milestone'
- [ ] For docs, I have run 'make docusaurus_start'
- [ ] For code, I have run 'make test_all'
- [ ] For configurations, I have update the documentation
- [ ] I added TODOs where applicable


adshmh (changes requested):
Thank you for the PR.
The direction looks good in general: avoids spamming endpoints with unnecessary quality check relays.

The main item to address is that the `QualityCheck` interface leaks QoS functionality out of the `qos` package and into the `Hydrator` IMO:
Summary:
- Expiry, etc. should be handled inside the QoS service, not by the Hydrator.
- Note: the above is the key reason why the `GetRequiredQualityChecks` method requires an endpoint address.

Details:
- A QoS service is expected to provide one or more `RequestQoSContext`s for a specific endpoint address.
- QoS Service should decide, internally and on its own, whether existing data on an endpoint is sufficient or more checks are needed.
- Hydrator should SOLELY supply the functionality to query each QoS service on what it needs on each endpoint, and return endpoints responses to QoS Service for processing.
- Hydrator should not contain any logic to check for and/or act on  whether an existing QoS result has expired: that is a responsibility of the QoS service.

qos/evm/state.go
```diff
@@ -1,27 +1,39 @@
...
 
 	"github.com/buildwithgrove/path/protocol"
 )
 
+// EndpointStoreConfig captures the modifiable settings of the EndpointStore.
+// This will enable `EndpointStore` to be used as part of QoS for other EVM-based
+// blockchains which may have different desired QoS properties.
+// e.g. different blockchains QoS instances could have different tolerance levels
+// for deviation from the current block height.
+type serviceStateConfig struct {
```
Olshansk, line 16: Do we need a `TODO_REFACTOR` to ensure that we have this logic mimiced in the other qos modules as well?

qos/evm/state.go
```diff
@@ -1,27 +1,39 @@
...
 import (
-	"fmt"
 	"sync"
 
 	"github.com/pokt-network/poktroll/pkg/polylog"
 
 	"github.com/buildwithgrove/path/protocol"
 )
 
+// EndpointStoreConfig captures the modifiable settings of the EndpointStore.
```
Olshansk, line 11: The struct is called `serviceStateConfig` so I believe there was some copy-pasta.

Please revisit the whole comment.

qos/evm/check.go
```diff
@@ -16,34 +19,63 @@ const (
 	idBlockNumberCheck
 )
 
-// EndpointStore provides the endpoint check generator required by
-// the gateway package to augment endpoints' quality data,
-// using synthetic service requests.
-var _ gateway.QoSEndpointCheckGenerator = &EndpointStore{}
+// endpointCheckName is a type for the names of the checks applied to an endpoint.
+type endpointCheckName string
```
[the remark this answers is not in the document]
Olshansk: Will this ever be an iota or an enum or just a general string?

qos/evm/check_blocknumber.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+var _ check = &endpointCheckBlockNumber{}
+
+const (
+	checkNameBlockNumber endpointCheckName = "block_number"
+	// methodBlockNumber is the JSON-RPC method for getting the latest block number.
+	// Reference: https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_blocknumber
+	methodBlockNumber = jsonrpc.Method("eth_blockNumber")
+	// TODO_IMPROVE: determine an appropriate interval for checking the block height.
```
Olshansk, line 17:
```suggestion
	// TODO_IMPROVE: determine an appropriate interval for checking the block height that
	// is a function of the chain's actual block time.
```

qos/evm/check_blocknumber.go
```diff
@@ -0,0 +1,63 @@
...
+	checkBlockNumberInterval = 60 * time.Second
+)
+
+var (
+	errNoBlockNumberObs      = fmt.Errorf("endpoint has not had an observation of its response to a %q request", methodBlockNumber)
+	errInvalidBlockNumberObs = fmt.Errorf("endpoint returned an invalid response to a %q request", methodBlockNumber)
+)
+
+// endpointCheckBlockNumber is a check that ensures the endpoint's block height is greater than the perceived block height.
+// It is used to ensure that the endpoint is not behind the chain.
```
Olshansk, line 27:
```suggestion
// It is used to ensure that the endpoint is not behind the perceived height of the chain.
```

qos/evm/check_blocknumber.go
```diff
@@ -0,0 +1,63 @@
...
+type endpointCheckBlockNumber struct {
+	// parsedBlockNumberResponse stores the result of processing the endpoint's response to an `eth_blockNumber` request.
```
Olshansk, line 29: #PUC for bot of these fields

qos/evm/check_chainid.go
```diff
@@ -0,0 +1,58 @@
...
+
+var (
+	errNoChainIDObs      = fmt.Errorf("endpoint has not had an observation of its response to a %q request", methodChainID)
+	errInvalidChainIDObs = fmt.Errorf("endpoint returned an invalid response to a %q request", methodChainID)
+)
+
+// endpointCheckChainID is a check that ensures the endpoint's chain ID is the same as the expected chain ID.
+// It is used to ensure that the endpoint is on the correct chain.
+type endpointCheckChainID struct {
+	// chainIDResponse stores the result of processing the endpoint's response to an `eth_chainId` request.
```
Olshansk, line 29: #PUC - same comment as elsewhere

qos/evm/check_empty.go
```diff
@@ -0,0 +1,41 @@
+package evm
+
+import (
+	"errors"
+)
+
+var _ check = &endpointCheckEmptyResponse{}
+
+// TODO_MVP(@commoddity): should we provide for a mechanism to un-sanction an endpoint that has returned an empty response?
```
Olshansk, line 9: I think we should remove this.

This is why we have session & permanent sanctions. 

In addition, restarts cancel permanent sessions.

I feel like we're getting into bikeshedding territory with this comment for now. If/when the time to revisit this comes up, I'm sure it'll float up on its own.

qos/evm/check_empty.go
```diff
@@ -0,0 +1,41 @@
...
+func (e *endpointCheckEmptyResponse) isValid(serviceState *ServiceState) error {
+	if e.hasReturnedEmptyResponse {
+		return errHasReturnedEmptyResponse
+	}
+	return nil
+}
+
+// shouldRun will always return false as returning an empty response disqualifies an endpoint
+// for the entire session and there is no requestContext to run for this check.
+// TODO_MVP(@commoddity): should we provide for a mechanism to un-sanction an endpoint that has returned an empty response?
```
Olshansk, line 38: See my comment above - my answer is no.

Open to pushback and/or feedback from @adshmh.

qos/evm/context.go
```diff
@@ -195,3 +197,10 @@ func preSelectedEndpoint(
 
 	return protocol.EndpointAddr(""), fmt.Errorf("singleEndpointSelector: endpoint %s not found in available endpoints", preSelectedEndpointAddr)
 }
+
+// setPreSelectedEndpointAddr assigns the endpoint address to be used for hydrator checks.
+// It is called to override the endpoint selection process with a specific endpoint.
+// and is used to enforce performing quality checks on a specific endpoint.
+func (rc *requestContext) setPreSelectedEndpointAddr(endpointAddr protocol.EndpointAddr) {
```
Olshansk, line 204: Can you add a TODO_IMPROVE to have a "golden fallback endpoint" that's configurable via env variables to PATH?

qos/evm/endpoint.go
```diff
@@ -1,100 +1,148 @@
...
-	// These endpoints are excluded from selection until service restart.
-	hasReturnedEmptyResponse bool
+	checks map[endpointCheckName]*evmQualityCheck
+}
 
-	// chainIDResponse stores the result of processing the endpoint's response to an `eth_chainId` request.
-	// It is nil if there has NOT been an observation of the endpoint's response to an `eth_chainId` request.
-	chainIDResponse *string
+// newEndpoint initializes a new endpoint with the checks that should be run for the endpoint.
+func newEndpoint(es *EndpointStore) endpoint {
```
Olshansk, line 18: Other than the naming, I love the new `/check_XXX.go` pattern!

qos/evm/endpoint.go
```diff
@@ -1,100 +1,148 @@
...
+	observedChainID := chainIDResponse.GetChainIdResponse()
+
+	e.checks[checkNameChainID].check = &endpointCheckChainID{
+		chainID:   &observedChainID,
+		expiresAt: time.Now().Add(checkChainIDInterval),
+	}
+}
+
+// applyBlockNumberObservation updates the block number check if a valid observation is provided.
+func (e *endpoint) applyBlockNumberObservation(blockNumberResponse *qosobservations.EVMBlockNumberResponse) {
```
Olshansk, line 113: I kind of feel like these functions should live in the corresponding `check_XXX.go` file.

I know the receiver is an `endpoint` but it does seem a bit weird to centralize it.

Not a strong opinion here but just a gut feeling.

Olshansk (changes requested):

PR #173, opened 2025-02-24
Add ai helpers and PR description generators

Olshansk:
## Summary
Add AI helper tools for generating PR descriptions and update Git configuration

### Primary Changes:
- Add `git_diff_copy_for_pr_description` make target to generate PR diffs in JSON format
- Add prompt template for AI-assisted PR description generation

### Secondary changes:
- Include new `ai_helpers.mk` makefile in the main `Makefile`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,7 @@ local/path/
 
 # Config Backup files
 *.backup
+
+# Codeweaver
+codeweaver/
+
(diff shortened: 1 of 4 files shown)


commoddity:
I think that the functionality in GDI would render including this in PATH redundant as we can do it from the CLI. IMO including too much stuff in PATH not related to its core functionality is not desirable. I also think it would be good to have a single "source of truth" for the PR generation. Just my thoughts on this but I do hope that once you use the CLI you will agree.

Issue #180, opened 2025-03-07
[QoS] First Version of EVM Archival Checks


fredteumer:
## Objective

Implement the first (simple & dynamic) Archival Check for PATH.

## Origin Document

Some Suppliers advertise themselves as containing the full ledger data. This may or may not be the case.

PATH needs a QoS observation implemented to "verify" that this (full ledger / archival) functionality is being provided and remove/sanction Suppliers that are not providing this service.

## Goals

- Implement a standard EVM Archival Check

### EVM Archival Check Implementation Details

Since not all Genesis blocks start at block `0x0`, the EVM check will be:
- Define the current block as `current_block`
- Run `eth_getBlockByNumber` on a random block between `current_block` / 2 and `current_block` / 10
- Verify that the response was a `non-error` code
- Take appropriate actions (i.e. permanent sanction) against failing nodes

## Deliverables

- [ ] A dynamic but simplistic EVM Archival check (outlined above).
- [ ] A new page in path.grove.city called "Archival QoS" that captures the details of how this is implemented and sets things up for other related documentation in the future

## Non-goals / Non-deliverables

- Any Non-EVM compatible archival checks
- Any other QoS Checks outside of the Archival Check

---

**Creator**: @fredteumer 
**Co-Owners**: @adshmh

fredteumer:
Fixed in #194 closing as there are more specific issues now to address.

PR #185, opened 2025-03-12
[DRAFT][WIP][DO NOT REVIEW] Archival Check Implementation Feedback

fredteumer:
## Summary

This was my attempt at implementing the Archival Check. It's unfinished, but it grew in complexity pretty quickly even with the assistance of AI. 

Here's my Claude conversation for reference: https://claude.ai/share/4f831439-ad07-435c-b274-b05ac6ab1f7e

## Feedback

My overall feedback and thinking going into this (as a `C` background guy - I don't know much about `go`) would be that the implementation would be as follows (max 3 files changed): 
- Define the check function (some equivalent to a `.h` file)
- Implement the function (equivalent to a `.c` file)
- Add one line in the `check.go` file to call the function

Per Claude (and maybe he and I were being dumb together) had me touching 5+ files with hundreds of lines of code. This is far too complex to be considered extensible if it's not a PEBCAK.

## Issue

- #180 

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

### DIFF
--- a/qos/evm/check.go
+++ b/qos/evm/check.go
@@ -1,6 +1,10 @@
 package evm
 
 import (
+	"crypto/rand"
+	"fmt"
+	"math/big"
+
 	"github.com/pokt-network/poktroll/pkg/polylog"
 
 	"github.com/buildwithgrove/path/gateway"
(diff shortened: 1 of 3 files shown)


Olshansk:
Closing this out (the PR is still preferable) as @commoddity picks up the work.

PR #194, opened 2025-03-26
[QoS][Config] feat: add EVM archival check and rework config service QoS file

commoddity:
## 🌿 Summary

Add EVM archival check and rework config service QoS file to support structured service configurations.

### 🛠️ Methodology
The archival check verifies that nodes can provide accurate historical blockchain data. Here's how it works:

1. Identify specific blockchain contracts that have been widely used since early in the chain's history.
2. The system selects a random historical block from the past (between when the contract was first deployed and what's considered "recent" - currently 128 blocks below latest block height for EVM).
3. PATH queries all endpoints in session for the contract's balance with `eth_getBalance` at this historical block.
4. When `n` endpoints (hardcoded to 5 but can be configurable) independently report the same balance value, this becomes our "ground truth" for that contract at that block.
5. Each endpoint is evaluated against this established truth - any endpoint that reports a different balance value is flagged as lacking proper archival data and will be filtering out by QoS.

### 🌱 Primary Changes:
- Implement EVM archival check with trie node ban and consensus mechanism
- Rework config/service_qos.go to use structured service configurations instead of maps
- Add archival check configurations for Ethereum, Polygon, Oasys, and XRPL EVM Testnet

### 🍃 Secondary changes:
- Add documentation for EVM archival checks in docusaurus/docs/learn/qos/evm_archival.md
- Update protobuf definitions to include EVMGetBalanceResponse for archival checks
- Improve error handling and validation in EVM QoS checks

## 💡 New TODOs 

- TODO_IMPROVE(@commoddity): Add a target similar to "make docs_update_gov_params_page" in poktroll that converts "config/service_qos.go" into markdown documentation. - Makefile
- TODO_QOS(@commoddity): Add archival check configurations for all EVM services. This includes hydrating the entire EVMArchivalCheckConfig struct. 
  - See the following archival check configurations as reference: - F00C (Ethereum) - F021 (Polygon) - F01C (Oasys) - F036 (XRPL EVM Testnet) - config/service_qos.go
- TODO_MVP(@commoddity): add archival check configuration for CometBFT. - qos/cometbft/qos.go
- TODO_IMPROVE(@commoddity): implement Allowance for block number check to allow blocks within a certain range of the current block number to still be considered valid to serve requests. - qos/evm/endpoint.go
- TODO_TECHDEBT(@commoddity): settle on a final value for this. - qos/evm/state_archival.go
- TODO_IMPROVE(@commoddity): set an expiry time for the archival state so that a new random block can be selected on an interval. - qos/evm/state_archival.go
- TODO_MVP(@commoddity): Use the Solana ServiceConfig for QoS - qos/solana/qos.go

## 🛠️ Type of change

## 🤯 Sanity Checklist


Olshansk (changes requested):
_tl;dr We should not tie `archival` to a specific call._

## Context ##

I saw this:

```proto
    // Response to `eth_getBlockByNumber` request
    // References:
    // * https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getblockbynumber
    EVMArchivalResponse archival_response = 4;
```

Note that archival just means "old data" and a bit arbitrary
- The types of requests vary from one chain to another   (it's not just one)
- The number of blocks that's considered archival is arbitrary based on the provider (i.e. gateway). E.g. Grove could make it 1000 blocks and Infura could decide it's 100. There is no hard definition/

See Alchemy's docs: https://docs.alchemy.com/docs/what-is-archive-data-on-ethereum

Screenshot 2025-03-27 at 12 06 06 AM

I also worked with ChatGPT on this:

```markdown
| Blockchain         | Archival Request Example                        | How Far Back Blocks Go                     | Purpose of Archival Request                                   | Providers and Their Archival Support                                           |
|--------------------|--------------------------------------------------|--------------------------------------------|----------------------------------------------------------------|--------------------------------------------------------------------------------|
| **EVM (Ethereum, BSC, etc.)** | `eth_getBalance` at block height           | Genesis block (full history)                | Retrieve balance at a specific block height                     | **QuickNode, Alchemy, Infura**: Typically retain full archival history, but default to latest 128 blocks for performance. |
|                    | `eth_call` with block number                     | Genesis block                               | Execute contract call at a given block height                   | **QuickNode**: Supports full archival history if requested.                     |
|                    | `eth_getStorageAt` at block height               | Genesis block                               | Retrieve contract storage at a past block height                 | **Alchemy**: Full archival support on premium plans.                            |
|                    | `eth_getLogs` with specific block range          | Genesis block                               | Query logs and events from past blocks                           | **Infura**: Archival support on premium plans.                                   |
| **Solana**          | `getAccountInfo` at a specific slot              | Typically a few hundred million slots       | Fetch historical account data                                    | **QuickNode**: Limited archival support; typically recent history only.         |
|                    | `getBlock` at specific slot                      | Usually retained for a few months           | Retrieve a full block from a specific slot                        | **Alchemy**: No full archival support.                                           |
|                    | `getConfirmedTransaction` at specific slot       | Limited, often a few months                  | Fetch a confirmed transaction at a given slot                     | **QuickNode**: Recent slots only.                                                |
| **CometBFT (Tendermint)** | `block` with specific height               | Pruned frequently (configurable)             | Fetch a specific block’s header and data                          | No major provider support; typically self-hosted.                                |
|                    | `block_results` at height                        | Pruned frequently (configurable)             | Retrieve transaction results for a given block height             | No standard archival support from common providers.                             |
| **Bitcoin (Full Node)** | `getblock` by block height                   | Genesis block                               | Retrieve block data from a specific height                        | **Blockstream**, **Chain API**: Full archival support.                          |
|                    | `getblockhash` by block height                   | Genesis block                               | Obtain the hash of a block at a specific height                    | **QuickNode**: Full archival support.                                             |
| **Cosmos SDK Chains** | `query txs` by block height                   | Pruned frequently (configurable)             | Query transactions included in a specific block                    | **Figment**, **Mintscan**: Typically latest state, not archival.                 |
| **Polkadot**        | `state_getStorage` at block height               | Full history (if archival mode)              | Fetch storage data at a specific block height                     | **OnFinality**, **SubQuery**: Archive nodes available on request.                 |
| **Tezos**           | `get /chains/main/blocks/<block>`                | Genesis block                               | Fetch a block and its metadata                                    | **TzStats**, **Baking Bad**: Full archival support.                             |
```


## Suggestion ##

1. Add `EVMArchivalMetadata`
2. Embed `EVMArchivalMetadata` in `EVMEndpointObservation` alongside `response_observation`
3. Refacor implementation so we can selectively apply the same archival logic to any request and make the "look back length" configurable by gateway.

```
// EVMArchivalMetadata stores information about the archival nature of a response.
message EVMArchivalMetadata {
  // Indicates whether the response is archival.
  bool is_archival = 1;
  // Number of blocks back that the archival data covers.
  uint64 archival_length = 2;
}

// EVMEndpointObservation stores a single observation from an endpoint servicing the protocol response.
// Example: A Pocket node on Shannon backed by an Ethereum data node servicing an `eth_getBlockNumber` request.
message EVMEndpointObservation {
  // Address of the endpoint handling the request (e.g., onchain address of a Pocket Morse/Shannon node)
  string endpoint_addr = 1;
  // Archival metadata, optional and present only if archival data is involved.
  EVMArchivalMetadata archival_metadata = 2;
  // Details of the response received from the endpoint
  oneof response_observation {
```

qos/evm/state.go
```diff
@@ -29,72 +31,164 @@ type ServiceState struct {
...
-		return fmt.Errorf("endpoint has block height %d, perceived block height is %d", blockNumber, perceivedBlockNumber)
+	s.archivalState.blockNumberHex = result
+	return result
+}
+
+// updateArchivalBalance checks for consensus and updates the archival balance if it hasn't been set yet.
+func (s *ServiceState) updateArchivalBalance(consensusThreshold int) {
+	for balance, count := range s.archivalState.balanceConsensus {
+		if count >= consensusThreshold {
+			s.archivalState.balance = balance
```
Olshansk, line 191: This actually breaks at the first balance that exceeds the threshold, but not necessarily the one with the highest count.

- If this is intentional, #PUC
- If this is not intentional, #PUC + TODO

qos/evm/state.go
```diff
@@ -29,72 +31,164 @@ type ServiceState struct {
...
+			break
+		}
 	}
+}
 
-	return nil
+// shouldPerformArchivalCheck returns a boolean indicating whether the archival check should be performs.
+// This depends on:
+//   - The archival check is enabled for the service
+//   - The archival block number to check the balance of has been set in the service state.
```
Olshansk: Please add a why/when to make it even clearer!

```suggestion
//   - The archival check is enabled for the service in the configs
//   - The archival block number to check the balance of has been set in the service state during ????
```

qos/evm/endpoint.go
```diff
@@ -73,28 +125,25 @@ func (e *endpoint) ApplyObservation(obs *qosobservations.EVMEndpointObservation)
...
-		// The endpoint returned an invalid response to an `eth_blockNumber` request.
-		// Explicitly set the parsedBlockNumberResponse to a zero value as the ParseUInt does not guarantee returning a 0 on all error cases.
-		if err != nil {
-			zero := uint64(0)
-			e.parsedBlockNumberResponse = &zero
+		e.parsedBlockNumberResponse = parseBlockNumberResponse(blockNumberResponse.GetBlockNumberResponse())
+		return true
+	}
+
+	if shouldPerformArchivalCheck {
```
Olshansk, line 136: Whenever I start seeing `if flagForFeature {` in more than one place in the codebase, it creates a bit of a 👃 and gets my 🕷️ senses tingling.

I don't have a better suggestion (yet), but we could potentially have a `ApplyObservaionArchival` function and call that from within this one.

Wdyt?

qos/evm/endpoint.go
```diff
@@ -33,34 +44,75 @@ type endpoint struct {
...
-	default:
-		return nil
 	}
+	if *e.chainIDResponse != chainID {
+		return fmt.Errorf("%s. expected: %s, got: %s", errInvalidChainIDObs, chainID, *e.chainIDResponse)
+	}
+	return nil
+}
+
+func (e *endpoint) validateBlockNumber(perceivedBlockNumber uint64) error {
```
Olshansk, line 73: godoc here and elsewhere

qos/evm/endpoint.go
```diff
@@ -6,15 +6,26 @@ import (
...
 var (
-	errNoChainIDObs             = fmt.Errorf("endpoint has not had an observation of its response to a %q request", methodChainID)
-	errInvalidChainIDObs        = fmt.Errorf("endpoint returned an invalid response to a %q request", methodChainID)
-	errNoBlockNumberObs         = fmt.Errorf("endpoint has not had an observation of its response to a %q request", methodBlockNumber)
-	errInvalidBlockNumberObs    = fmt.Errorf("endpoint returned an invalid response to a %q request", methodBlockNumber)
+	// empty response errors
 	errHasReturnedEmptyResponse = errors.New("endpoint is invalid: history of empty responses")
+
+	// chainID check errors
+	errNoChainIDObs      = fmt.Errorf("endpoint has not had an observation of its response to a %q request", methodChainID)
```
Olshansk, line 18: Seems like we could move a few things to a shared string like `errNoObservationStr` and `errInvalidObsStr` and reuse it.

qos/evm/check.go
```diff
@@ -24,11 +27,18 @@ var _ gateway.QoSEndpointCheckGenerator = &EndpointStore{}
 // TODO_IMPROVE(@commoddity): implement QoS check expiry functionality and use protocol.EndpointAddr
 // to filter out checks for any endpoint which has acurrently valid QoS data point.
 func (es *EndpointStore) GetRequiredQualityChecks(_ protocol.EndpointAddr) []gateway.RequestQoSContext {
-	return []gateway.RequestQoSContext{
+	qualityChecks := []gateway.RequestQoSContext{
```
Olshansk, line 30: I find it a bit weird that this is called `qualityChecks` but we return a list of `QoSContexts`.

Not modifying because I want to see if this was intentional before modifying it.

qos/evm/check.go
```diff
@@ -50,19 +60,53 @@ func getEndpointCheck(
...
+// withArchivalBlockCheck updates the request context to make an EVM JSON-RPC eth_getBalance request with a random archival block number.
+// eg. '{"jsonrpc":"2.0","id":1,"method":"eth_getBalance","params":["0x28C6c06298d514Db089934071355E5743bf21d60", "0xe71e1d"]}'
+func withArchivalBlockCheck(requestCtx *requestContext) {
+	// Get the archival block number from the endpoint store.
+	archivalCheckConfig := requestCtx.endpointStore.serviceState.archivalCheckConfig
+	// Get the current state of the archival check.
+	serviceArchivalState := requestCtx.endpointStore.serviceState.archivalState
+
+	requestCtx.jsonrpcReq = buildJSONRPCReq(
+		idArchivalBlockCheck,
```
Olshansk, line 88: Wdyt of `archivalBlockCheckId` instead of `idArchivalBlockCheck`?

qos/evm/check.go
```diff
@@ -50,19 +60,53 @@ func getEndpointCheck(
...
+		methodGetBalance,
+		// Pass params in this order, eg. "params":["0x28C6c06298d514Db089934071355E5743bf21d60", "0xe71e1d"]
+		// Reference: https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getbalance
+		archivalCheckConfig.ContractAddress,
+		serviceArchivalState.blockNumberHex,
```
Olshansk, line 93: Why aren't we using `archivalCheckConfig.perceivedBlockNumber`  or `archivalCheckConfig.evmArchivalState.blockNumberHex` here?

Need to #PUC so its immediately evident to anyone, even if they only review this small part of the codebase.

qos/evm/check.go
```diff
@@ -50,19 +60,53 @@ func getEndpointCheck(
...
+	)
+
+	// Set the archival balance check flag to true.
+	// This is used to ensure that only hydrator requests for the archival block number are used
```
Olshansk, line 97: I read this a couple of times, but it's still unclear to me why we need it.

Can we go without it? If not, can you try to update the comment so its clearer?

Olshansk (changes requested):
@commoddity Let me start off by saying that this is great work and I found the review, though not easy, very rewarding!

I pushed some changes in [`a8f6441` (#194)](https://github.com/buildwithgrove/path/pull/194/commits/a8f6441245c2942489f5fef81727b87b034e6c68), but the key asks/questions are in the comments.

I think it's very critical we get this right because it's a VERY important feature and will set the foundation for a lot more to come.

Whatever refactoring comes next, we're all aligned (cc @adshmh) that we need to get this in first.

As you look at my comments, note that I was jumping around and building context throughout. There's a chance some may not make as much sense, so it's possible that they were earlier in the review process.

If anything is ultra confusing, just ping on discord rather than waiting for a response here.

@adshmh Please see my changes / comments as you review, and let's prioritize review #2 of this as soon as @commoddity says it's ready again!

qos/evm/state.go
```diff
@@ -29,72 +31,164 @@ type ServiceState struct {
...
-	if err := validateEndpointBlockNumber(endpoint, s.perceivedBlockNumber); err != nil {
+	if err := endpoint.validateChainID(s.chainID); err != nil {
 		return err
 	}
-
+	if err := endpoint.validateBlockNumber(s.perceivedBlockNumber); err != nil {
+		return err
+	}
+	if s.shouldPerformArchivalCheck() {
+		if err := endpoint.validateArchivalCheck(s.archivalState.balance, endpointAddr); err != nil {
```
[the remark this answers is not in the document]
adshmh: Alternative:
- keep the `endpoint` struct as a container
- have a method/entity under `State` that performs the validation of an endpoint with regards to current archival state.

Advantage: consolidate Archival-related logic to a single struct under the `State` struct.

qos/evm/state.go
```diff
@@ -29,72 +31,186 @@ type ServiceState struct {
...
-		logger.With("endpoint_block_number", blockNumber).Info().Msg("Updating latest block height")
+		// Attempt to retrieve the archival balance from the endpoint.
+		balance, err := endpoint.getArchivalBalance()
+		if err != nil {
+			logger.Info().Err(err).Msg("Skipping endpoint without archival balance")
+			continue
+		}
+
+		// Only count non-empty balances toward consensus.
+		if s.archivalState.balance == "" && balance != "" {
```
adshmh, line 138: The logic in this method is becoming too complex to follow (granted it was already a good refactor candidate).

WDYT of moving all Archival related logic into a separate struct under `State`?
The "archival check" struct can focus solely on the `archivalState`.
This will also consolidate all the "archival" methods into a single struct.

qos/evm/check.go
adshmh: This is a good call out, but the required refactor may over extend the scope of this PR.

- Quality checks should also be determined by the ServiceState (which has access to EndpointStore)
- I suggest adding a TODO_MVP/TODO_TECHDEBT (will most likely be addressed by the refactor work).

qos/evm/check.go
```diff
@@ -24,11 +27,21 @@ var _ gateway.QoSEndpointCheckGenerator = &EndpointStore{}
...
 		getEndpointCheck(es.logger, es, withBlockHeightCheck),
-		// TODO_FUTURE: add an archival endpoint check.
 	}
+
+	// If all of the following are true:
+	//  - Service is configured to perform an archival check
+	//  - Has calculated an expected archival balance (balance at a specific block number for a specific address for a particular chain)
+	// Then:
+	// - Add the archival check to the list of qos checks to perform on every hydrator run
+	if es.serviceState.shouldPerformArchivalCheck() {
```
adshmh, line 40: WDYT of the following to make the responsibilities more clear without over extending the PR's scope:
- The ServiceState can return a (JSONRPC request, bool) pair, the latter indicating whether an archival check should be added.
- The code here simply updates the ID of the request (if necessary).

qos/evm/check.go
```diff
@@ -50,19 +63,55 @@ func getEndpointCheck(
...
+		// Pass params in this order, eg. "params":["0x28C6c06298d514Db089934071355E5743bf21d60", "0xe71e1d"]
+		// Reference: https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getbalance
+		archivalCheckConfig.ContractAddress,
+		serviceArchivalState.blockNumberHex,
+	)
+
+	// Set the archival balance check flag to true.
+	// This is used to ensure that only hydrator requests for the archival block number are used
+	// to update QoS data on whether endpoints are able to service archival requests.
+	requestCtx.archivalBalanceCheck = true
```
adshmh, line 99: Could you help clarify why this may be needed?
Suggestion:
- RequestContext should not know about the source of a request (user/hydrator)
- The ArchivalCheck component in the ServiceState can choose to ignore an observation if it is not helpful for establishing Archival State of one or more endpoints.

qos/evm/check.go
```diff
@@ -50,19 +63,55 @@ func getEndpointCheck(
...
+}
+
+func buildJSONRPCReq(id int, method jsonrpc.Method, params ...any) jsonrpc.Request {
+	request := jsonrpc.Request{
 		JSONRPC: jsonrpc.Version2,
 		ID:      jsonrpc.IDFromInt(id),
 		Method:  method,
 	}
+
+	if len(params) > 0 {
```
adshmh, line 109: This looks like jsonrpc functionality that belongs in the `qos/jsonrpc` package

adshmh (changes requested):
Great work! 
A few comments to address, mainly:
It would be great if we could consolidate all archival-related logic into a separate component/struct under `ServiceState`.

proto/path/qos/evm.proto
```diff
@@ -144,18 +144,23 @@ message EVMEndpointObservation {
     // * Chain IDs: https://chainlist.org
     EVMBlockNumberResponse block_number_response = 3;
 
+    // Response to `eth_getBlockByNumber` request
+    // References:
+    // * https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getblockbynumber
+    EVMArchivalResponse archival_response = 4;
```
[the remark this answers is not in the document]
commoddity: This is outdated and I should have updated it; archival checks use `eth_getBalance` only.

This will be a recurring theme in the review but I do think there is a need to distinguish between:
- regular `eth_getBalance` requests (which can be for any block) 
- archival `eth_getBalance` requests (which check a very specific block and must return a very specific balance for that block to be valid)

The main reason is that QoS will be expecting the correct balance for the correct block and so requests for other blocks will throw off the results. 

I'm not opposed to refactoring the **way** in which this is handled as I don't love the current solution of putting a bool flag on the request context. However, I do believe it is important that there is some (reliable) way to distinguish archival checks from other `eth_getBalance` requests.

CC @Olshansk for context.

protocol/morse/context.go
```diff
@@ -41,7 +41,7 @@ func (rc *requestContext) HandleServiceRequest(payload protocol.Payload) (protoc
 	// TODO_IMPROVE(@adshmh): use the same pattern for hydrated loggers in any packages with large number of logger fields.
 	hydratedLogger := rc.getHydratedLogger("HandleServiceRequest")
 
-	if rc.selectedEndpoint == nil {
+	if rc.selectedEndpoint.IsEmpty() {
```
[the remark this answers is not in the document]
commoddity: No problem, this was done in a previous version of the PR where the endpoint was not a pointer. I've reverted to the nil check.

protocol/morse/endpoint.go
```diff
@@ -30,6 +30,10 @@ type endpoint struct {
 	app app
 }
 
+func (e endpoint) IsEmpty() bool {
```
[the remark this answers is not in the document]
commoddity: Ditto; removed the method.

qos/evm/state.go
```diff
@@ -29,72 +31,164 @@ type ServiceState struct {
...
-	if blockNumber < perceivedBlockNumber {
-		return fmt.Errorf("endpoint has block height %d, perceived block height is %d", blockNumber, perceivedBlockNumber)
+	s.archivalState.blockNumberHex = result
+	return result
+}
+
+// updateArchivalBalance checks for consensus and updates the archival balance if it hasn't been set yet.
+func (s *ServiceState) updateArchivalBalance(consensusThreshold int) {
+	for balance, count := range s.archivalState.balanceConsensus {
+		if count >= consensusThreshold {
```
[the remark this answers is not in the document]
commoddity: Yes this is the intention; the balance for the block number is considered correct when `n` number of endpoints agree on the balance for the block; as soon as they do, it's considered the source of truth. 

Currently this is set to 5 but could be adjusted (I don't think there's much value in making it configurable, though perhaps the hardcoded value could be increased; that said I think the chances that >5 random endpoints will agree on an incorrect balance for a random old block is negligible).

qos/evm/store.go
```diff
@@ -77,7 +77,9 @@ func (es *EndpointStore) Select(availableEndpoints []protocol.EndpointAddr) (pro
 	).Info().Msg("filtered endpoints")
 
 	// TODO_FUTURE: consider ranking filtered endpoints, e.g. based on latency, rather than randomization.
-	return filteredEndpointsAddr[rand.Intn(len(filteredEndpointsAddr))], nil
+	// return filteredEndpointsAddr[rand.Intn(len(filteredEndpointsAddr))], nil
```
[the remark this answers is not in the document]
commoddity: It's been updated.

qos/jsonrpc/id.go
```diff
@@ -31,6 +32,17 @@ func (id ID) String() string {
 	return fmt.Sprintf("%d", id.intID)
 }
 
+func (id ID) Int() int {
+	if id.intID != 0 {
+		return id.intID
+	}
+	parsed, err := strconv.Atoi(id.strID)
```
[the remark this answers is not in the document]
commoddity: Good catch; this was left over from a previous iteration of the PR and is no longer needed. I've removed it.

qos/evm/check.go
commoddity: I have refactored to use this suggestion; I like it.

qos/evm/response.go
```diff
@@ -24,31 +24,28 @@ var (
...
+//   - eth_getBalance (if rc. archivalBalanceCheck is true)
 //   - any empty response, regardless of method
-func unmarshalResponse(
-	logger polylog.Logger,
-	jsonrpcReq jsonrpc.Request,
-	data []byte,
-) (
-	response, error,
-) {
+func (rc *requestContext) unmarshalResponse(data []byte) (response, error) {
```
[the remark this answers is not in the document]
commoddity: Updated to remove the need to make `unmarshalResponse` a method.

proto/path/qos/evm.proto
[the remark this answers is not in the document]
commoddity: After some consideration and implementation I have come up with a solution that removes the need to the various changes I think you were objecting to.

The new implementation works as follows:
- All `eth_getBalance` use the custom `eth_getBalance` unmarshaler; there is not more explicit definition marking them as archival at this stage.
- The custom unmarshaler utilizes a new `getBlockNumberFromRequest` to get the block number that the balance was requested for. This block number hex is passed to the `qosobservations.EVMGetBalanceResponse`.
- When observations are processed for these requests, the `ApplyObservation` method checks if the `qosobservations.EVMGetBalanceResponse`'s `blockNumber` field matches the block number hex that the service state expects to be checking to determine the archival balance. 
- If the block number matches then it applies the observation from the `eth_getBalance` request.

This implementation reduces complexity while still preserving the key requirement that only `eth_getBalance` requests for the archival check block height are applied to the EVM QoS archival validation check for endpoints.

CC @adshmh @Olshansk

qos/evm/check.go
commoddity: No longer needed. Has been removed.

qos/evm/check.go
commoddity: Also removed as this file has been refactored according to @adshmh 's suggestion here: https://github.com/buildwithgrove/path/pull/194#discussion_r2026804634

IMO it was a great suggestion and improvement.

qos/evm/check.go
commoddity: My preference is to structure enum names in the order `<type><specific value>` so I suggest leaving it unchanged.

qos/evm/check.go
commoddity: Agreed; additionally I also have a (new) pending PR to implement expiry functionality for QoS https://github.com/buildwithgrove/path/pull/200 (which will likely change as there are a few PRs preceding it). This PR updates the checks functionality as well so when we get to it we can revisit this issue.

protocol/morse/app.go
```diff
@@ -13,6 +13,10 @@ type app struct {
 	aat provider.PocketAAT
 }
 
+func (a app) IsEmpty() bool {
```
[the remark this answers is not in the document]
commoddity: It's not relevant anymore and has been removed.

proto/path/qos/evm.proto
```diff
@@ -144,18 +144,23 @@ message EVMEndpointObservation {
     // * Chain IDs: https://chainlist.org
     EVMBlockNumberResponse block_number_response = 3;
 
+    // Response to `eth_getBlockByNumber` request
```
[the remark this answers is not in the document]
commoddity: Ditto.

proto/path/qos/evm.proto
```diff
@@ -195,6 +200,20 @@ message EVMBlockNumberResponse {
   optional EVMResponseValidationError response_validation_error = 3 [(metadata.semantic_meaning) = "Validity failure type"];
 }
 
+// EVMArchivalResponse stores the response to an `eth_getBlockByNumber` request
+// https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getblockbynumber
+message EVMArchivalResponse {
+  // The HTTP status code received from the endpoint
+  int32 http_status_code = 1;
+
+  // The balance value at the archival block number
```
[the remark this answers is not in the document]
commoddity: Yes to both.

qos/cometbft/config.go
```diff
@@ -0,0 +1,29 @@
+package cometbft
```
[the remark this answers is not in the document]
commoddity: I think by definition we need different things in different services. Especially as we start to add more diverse kinds of services I think it's important to allow different services to have different configuration options.

I did attempt to try and make it as universal as possible but IMO the tradeoff in universality is not worth the number of redundant and unused fields we will wind up with.

tl;dr I strongly feel we should keep it as it is: concrete structs in `qos/<service>` package, type casting in `cmd/qos.go` file.

config/service_qos.go
```diff
@@ -1,177 +1,306 @@
...
-	for k := range shannonCometBFTChainIDs {
-		ServiceQoSTypes[k] = ServiceIDCometBFT
-	}
-	for k, v := range morseQoSTypes {
-		ServiceQoSTypes[k] = v
-	}
-	for k := range morseEVMChainIDs {
-		ServiceQoSTypes[k] = ServiceIDEVM
-	}
+type ServiceConfig interface {
```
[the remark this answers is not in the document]
commoddity: IMO the VirtualMachine terminology is confusing and doesn't relate to the rest of the repo in terms of naming.

I would prefer to leave this definition in the file where it's used but not strongly attached. To me it just seem to make the most sense there. 

And as far as making it an enum, I did initially have it set up this way but it led to import cycles once I moved the services to their own packages and so it was not worth making it a custom type vs a normal string.

config/service_qos.go
```diff
@@ -1,177 +1,306 @@
...
+	shannonServices: shannonServices,
+	morseServices:   morseServices,
 }
 
-// All CometBFT Shannon Service IDs and their corresponding CometBFT chain IDs.
-var shannonCometBFTChainIDs = map[protocol.ServiceID]string{
-	"pocket-beta-rpc": "pocket-beta",          // Pocket Beta Testnet
-	"cometbft":        defaultCometBFTChainID, // Cosmos Hub
-}
+const (
```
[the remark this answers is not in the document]
commoddity: I get where you're coming from but given that every other value is defined in this file IMO it's confusing to have to jump to EVM package and back to see what the value is.

config/service_qos.go
```diff
@@ -1,177 +1,306 @@
...
+	evm.ServiceConfig{
+		ServiceID:  "F00B",   // Celo
+		EVMChainID: "0xa4ec", // (42220)
+	},
+	evm.ServiceConfig{
+		ServiceID:  "F00C",            // Ethereum
+		EVMChainID: defaultEVMChainID, // (1)
+		ArchivalCheckConfig: evm.EVMArchivalCheckConfig{
+			Enabled: true,
+			// https://etherscan.io/address/0x28C6c06298d514Db089934071355E5743bf21d60
```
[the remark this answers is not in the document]
commoddity: > #PUC how/why this address was selected.

IMO writing this for every contract is going to add a lot of noise with no real benefit (it's the same rationale for every contract address: frequent transactions, large balance).

Instead I suggest updating the comment for the `EVMArchivalCheckConfig` struct with the general principle. WDYT?

eg.
```go
// EVMArchivalCheckConfig is the configuration for the archival check.
//
// The basic methodology is:
//  1. Select a `ContractAddress` for the chain with a frequent transaction volume and large balance.
//  2. Determine its starting block height (`ContractStartBlock`).
//  3. Set a `Threshold` for how many blocks below the current block number are considered "archival" data.
//
// With all of this data, the QoS implementation can select a random block number to check using `eth_getBalance`.
type EVMArchivalCheckConfig struct {
	Enabled            bool   // Whether to require an archival check for the service.
	Threshold          uint64 // The number of blocks below the current block number to be considered "archival" data
	ContractAddress    string // The address of the contract to check for the archival balance.
	ContractStartBlock uint64 // The start block of the contract address (ie. when it first had a balance)
}
```

qos/evm/response.go
```diff
@@ -58,23 +55,30 @@ func unmarshalResponse(
...
-	if err := jsonrpcResponse.Validate(jsonrpcReq.ID); err != nil {
-		return getGenericJSONRPCErrResponse(logger, jsonrpcReq.ID, data, err), err
+	if err := jsonrpcResponse.Validate(rc.jsonrpcReq.ID); err != nil {
+		return getGenericJSONRPCErrResponse(rc.logger, rc.jsonrpcReq.ID, data, err), err
 	}
 
 	// We intentionally skip checking whether the JSONRPC response indicates an error.
 	// This allows the method-specific handler to determine how to respond to the user.
 
+	// Filter out 'eth_getBalance' requests that are not performed as part of the archival check.
```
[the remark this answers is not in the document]
commoddity: Agreed; it's been refactored out.

See: https://github.com/buildwithgrove/path/pull/194#discussion_r2027566219

qos/jsonrpc/params.go
```diff
@@ -66,3 +70,14 @@ func (p *Params) UnmarshalJSON(data []byte) error {
 func (p Params) IsEmpty() bool {
 	return len(p.rawMessage) == 0
 }
+
+// Slice returns the params as a slice of any.
+// As the "params" field may be of any type, according to the JSON-RPC spec,
```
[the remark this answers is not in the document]
commoddity: Slice is no longer needed; removed in favour of a tightly scoped method in:
https://github.com/buildwithgrove/path/blob/evm-archival-check/qos/evm/response_getbalance.go#L130

qos/jsonrpc/params.go
```diff
@@ -26,6 +26,10 @@ type Params struct {
 	rawMessage json.RawMessage
 }
 
+func NewParams(rawMessage json.RawMessage) Params {
```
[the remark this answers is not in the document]
commoddity: No longer needed; `NewParams` has been removed.

qos/evm/state.go
```diff
@@ -29,72 +31,164 @@ type ServiceState struct {
...
 	// https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_blocknumber
 	perceivedBlockNumber uint64
+
+	// archivalCheckConfig contains all configurable values for an EVM archival check.
+	archivalCheckConfig EVMArchivalCheckConfig
+	// archivalState contains the current state of the EVM archival check.
+	archivalState evmArchivalState
+}
+
+// consensusThreshold is the number of endpoints that must agree on the archival balance for the randomly
```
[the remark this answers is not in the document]
commoddity: This is done. See:
https://github.com/buildwithgrove/path/blob/evm-archival-check/qos/evm/state_archival.go

qos/evm/state.go
```diff
@@ -29,72 +31,164 @@ type ServiceState struct {
...
+type evmArchivalState struct {
+	// blockNumberHex is a randomly selected block number from which to check the balance of the contract.
+	blockNumberHex string
+	// balance is the balance of the contract at the block number specified in `blockNumberHex`.
+	balance string
+	// balanceConsensus is a map of balances and the number of endpoints that reported them.
+	balanceConsensus map[string]int
+	// refreshAt is the time at which the archival state should be refreshed.
+	// If it has passed, the archival state should be refreshed by randomly selecting a new block number.
+	// TODO_IMPROVE(@commoddity): Implement a refresh mechanism to calculate a new archival block number
```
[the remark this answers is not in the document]
commoddity: > Why? I'm all for improvements, but don't see how this is better than the random block for this specific use case.

Currently the block number is set when PATH first starts up and - once set- will remain set as that value until PATH restarts again.

My idea was to have this re-select a new random block say every hour or two.

Not sure it's needed right now but could be a nice improvement for the future. I'll leave a TODO_IMPROVE.

qos/evm/state.go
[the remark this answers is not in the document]
commoddity: I've implemented elements of both suggestions to improve the clarity and cleanliness of ValidateEndpoint.

```go
// ValidateEndpoint returns an error if the supplied endpoint is not valid based on the perceived state of the EVM blockchain.
//
// It returns an error if:
// - The endpoint has not returned an empty response to a `eth_getBalance` request.
// - The endpoint's response to an `eth_chainId` request is not the expected chain ID.
// - The endpoint's response to an `eth_blockNumber` request is greater than the perceived block number.
// - The endpoint has not returned an archival balance for the perceived block number.
func (s *ServiceState) ValidateEndpoint(endpoint endpoint, endpointAddr protocol.EndpointAddr) error {
	s.serviceStateLock.RLock()
	defer s.serviceStateLock.RUnlock()

	// Ensure the response is not empty.
	if err := endpoint.validateEmptyResponse(); err != nil {
		return err
	}
	// Ensure the chain ID is valid.
	if err := endpoint.validateChainID(s.chainID); err != nil {
		return err
	}
	// Ensure the service state's perceived block number is not ahead of the endpoint's block number.
	if err := endpoint.validateBlockNumber(s.perceivedBlockNumber); err != nil {
		return err
	}
	// Ensure the endpoint has returned an archival balance for the perceived block number.
	// If the service does not require an archival check, this will always return a nil error.
	if err := endpoint.validateArchivalCheck(s.archivalState); err != nil {
		return err
	}

	return nil
}
```

qos/evm/state.go
```diff
@@ -29,72 +31,164 @@ type ServiceState struct {
...
 
 	return nil
 }
 
-// validateEndpointBlockNumber validates the supplied endpoint against the supplied perceived block number for the EVM blockchain.
-func validateEndpointBlockNumber(endpoint endpoint, perceivedBlockNumber uint64) error {
-	blockNumber, err := endpoint.GetBlockNumber()
-	if err != nil {
-		return err
+// assignArchivalBlockNumber returns a random archival block number based on the perceived block number.
```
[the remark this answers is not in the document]
commoddity: this is done.

qos/evm/state.go
```diff
@@ -29,72 +31,164 @@ type ServiceState struct {
...
+			s.archivalState.balanceConsensus[balance]++
+		}
+	}
+
+	// Update archival block number and archival balance only if not yet set.
+	if s.perceivedBlockNumber != 0 && s.archivalState.blockNumberHex == "" {
+		s.assignArchivalBlockNumber()
+	}
+	if s.archivalState.balance == "" {
+		s.updateArchivalBalance(consensusThreshold)
```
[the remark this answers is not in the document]
commoddity: Yup, this is now refactored and much improved IMO:
- https://github.com/buildwithgrove/path/blob/evm-archival-check/qos/evm/state.go
- https://github.com/buildwithgrove/path/blob/evm-archival-check/qos/evm/state_archival.go

qos/evm/state.go
```diff
@@ -29,72 +31,164 @@ type ServiceState struct {
...
+
+	var result string
+	if s.perceivedBlockNumber <= archivalThreshold {
+		result = blockNumberToHex(0)
+	} else {
+		maxBlockNumber := s.perceivedBlockNumber - archivalThreshold
+		if maxBlockNumber < minArchivalBlock {
+			result = blockNumberToHex(minArchivalBlock)
+		} else {
+			r := rand.New(rand.NewSource(time.Now().UnixNano()))
```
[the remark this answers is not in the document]
commoddity: IMO we want randomness here as the point is make it quite non-deterministic. I've done a lot of manual testing here and the consensus mechanism works great so would rather not change it.

qos/evm/state.go
commoddity: It's not checking balance but the number of endpoints that agree on the balance at the specific block height for the specific contract.

qos/evm/endpoint.go
commoddity: this has been fixed.

qos/evm/endpoint.go
```diff
@@ -94,29 +97,44 @@ func (e endpoint) getBlockNumber() (uint64, error) {
 	return *e.parsedBlockNumberResponse, nil
 }
 
-func (e *endpoint) validateArchivalCheck(archivalBalance string, endpointAddr protocol.EndpointAddr) error {
-	if e.archivalBalance == "" {
+// validateArchivalCheck validates that the endpoint has returned an archival balance for the perceived block number.
+
+// If the archival check is not enabled for the service, this will always return a nil error.
+func (e *endpoint) validateArchivalCheck(archivalState archivalState) error {
```
adshmh, line 103: Not a blocker: a TODO that we will need to move these checks to the ServiceState struct rather than passing the state to each endpoint.

qos/evm/endpoint.go
```diff
@@ -94,29 +97,44 @@ func (e endpoint) getBlockNumber() (uint64, error) {
...
+//
+// For archival balance observations:
+// - Only updates the archival balance if the balance was observed at the specified archival block height
+// - This ensures accurate historical balance validation at the specific block number
+//
 // TODO_TECHDEBT(@adshmh): add a method to distinguish the following two scenarios:
 //   - an endpoint that returned in invalid response.
 //   - an endpoint with no/incomplete observations.
-func (e *endpoint) ApplyObservation(obs *qosobservations.EVMEndpointObservation, shouldPerformArchivalCheck bool) bool {
+func (e *endpoint) ApplyObservation(obs *qosobservations.EVMEndpointObservation, archivalBlockHeight string) bool {
```
adshmh, line 137: Ditto here: not a blocker, but a TODO that these should be moved to ServiceState eventually, to remove the need for passing around multiple data items.

qos/evm/qos.go
```diff
@@ -14,9 +14,18 @@ func NewQoSInstance(logger polylog.Logger, config ServiceConfig) *QoS {
...
 
 	serviceState := &ServiceState{
-		logger:              logger,
-		chainID:             evmChainID,
-		archivalCheckConfig: config.GetEVMArchivalCheckConfig(),
+		logger:  logger,
+		chainID: evmChainID,
+	}
+
+	if archivalCheckConfig, enabled := config.GetEVMArchivalCheckConfig(); enabled {
```
adshmh, line 21: Not a blocker: not sure why running archival checks is optional: 
IMO, if not configured we should:
1. Use a set of default values.
2. Log a warning.

qos/evm/state.go
```diff
@@ -127,90 +95,20 @@ func (s *ServiceState) UpdateFromEndpoints(updatedEndpoints map[protocol.Endpoin
...
-		s.assignArchivalBlockNumber()
-	}
-	if s.archivalState.balance == "" {
-		s.updateArchivalBalance(archivalConsensusThreshold)
-	}
+	// Update the archival state based on the perceived block number
+	// This handles:
+	// 1. Calculating a random archival block number if needed
+	// 2. Determining consensus on the expected balance at that block number
+	s.archivalState.updateArchivalState(s.perceivedBlockNumber)
```
adshmh, line 111: Not a blocker, as I am not 100% sure it is possible: but making a single call to the archivalState and passing it all data that is relevant will make the code much easier to read.

qos/evm/state_archival.go
```diff
@@ -0,0 +1,190 @@
...
+// TODO_TECHDEBT(@commoddity): settle on a final value for this.
+const archivalConsensusThreshold = 5
+
+// The archival check verifies that nodes can provide accurate historical blockchain data. Here's how it works:
+//   - Identify specific blockchain contracts that have been widely used since early in the chain's history.
+//   - The system selects a random historical block from the past (between when the contract was first deployed and what's considered "recent");
+//   - PATH queries all endpoints in session for the contract's balance with eth_getBalance at this historical block.
+//   - When <archivalConsensusThreshold> endpoints independently report the same balance value, this becomes our "source of truth" for that contract at that block.
+//   - Each endpoint is evaluated against this established truth - any endpoint that reports a different balance value is flagged as lacking proper archival data and will be filtered out by QoS.
+type archivalState struct {
```
adshmh, line 24: Nice!

qos/evm/state_archival.go
```diff
@@ -0,0 +1,190 @@
...
+	a.updateConsensusMap(balance)
+	return true
+}
+
+// updateConsensusMap updates the balance consensus map to determine
+// the expected archival balance at the archival block number.
+//
+// Once <archivalConsensusThreshold> endpoints report the same balance,
+// the balance is set as the expected archival balance.
+func (a *archivalState) updateConsensusMap(balance string) {
```
adshmh, line 87: `Map` in the method leaks details the caller should not know about. `updateBalanceConsensus` (or similar) may be a better name here.

qos/evm/state_archival.go
```diff
@@ -0,0 +1,190 @@
...
+	if a.expectedBalance == "" && balance != "" {
+		a.balanceConsensus[balance]++
+	}
+}
+
+// updateArchivalState updates the archival state based on the perceived block number.
+// This handles calculating the archival block number if needed and updating the
+// consensus-based archival balance.
+func (a *archivalState) updateArchivalState(perceivedBlockNumber uint64) {
```
adshmh, line 96: If possible, merging the multiple method calls into a single flow/method would make the code much easier to read/follow: e.g. consume a `getPerceivedBlockNumber` interface and merge this method with the other update method(s).

qos/evm/state_archival.go
```diff
@@ -0,0 +1,190 @@
...
+func blockNumberToHex(blockNumber uint64) string {
+	return fmt.Sprintf("0x%x", blockNumber)
+}
+
+// getArchivalCheckRequest returns a JSONRPC request to check the balance of:
+//   - the contract specified in `a.archivalCheckConfig.ContractAddress`
+//   - at the block number specified in `a.blockNumberHex`
+//
+// It returns false if the archival check is not enabled for the service or if the block number has not been set.
+func (a *archivalState) getArchivalCheckRequest() (jsonrpc.Request, bool) {
```
adshmh, line 175: Moving this to the checks.go could help show its role better (not ideal as it splits a structs methods, but in this specific case can improve readability)

qos/evm/state_archival.go
```diff
@@ -0,0 +1,190 @@
...
+		a.updateArchivalBalance(archivalConsensusThreshold)
+	}
+}
+
+// calculateArchivalBlockNumber determines a random archival block number based on the perceived block number.
+// The function applies the following logic:
+//   - If perceived block is below threshold, returns block 0
+//   - Otherwise, calculates a random block between minArchivalBlock and (perceivedBlockNumber - threshold)
+//   - Ensures the returned block number is never below the contract start block
+func (a *archivalState) calculateArchivalBlockNumber(perceivedBlockNumber uint64) string {
```
adshmh, line 122: Overall, moving most of the code to this file is a great improvement.
Another iteration (or a TODO_IMPROVE) can organize everything into a single, clear flow of data/events.

qos/jsonrpc/request.go
```diff
@@ -23,6 +23,32 @@ type Request struct {
...
+// NewRequest constructs a new JSONRPC request from the given ID, method, and optional parameters.
+//
+// For example:
+// id - 1
+// method - eth_getBalance
+// params - ["0x28C6c06298d514Db089934071355E5743bf21d60", "0xe71e1d"]
+//
+// The above request would be serialized as:
+// {"jsonrpc":"2.0","id":1,"method":"eth_getBalance","params":["0x28C6c06298d514Db089934071355E5743bf21d60", "0xe71e1d"]}
+func NewRequest(id int, method Method, params ...any) Request {
```
adshmh, line 35: Please reconsider this (specifically use of `any`):
- All fields of `Request` are already public: `NewRequest` does not add any functionality and is not GO idiomatic.
- A helper method to build `Params` from a typed set of values, e.g. `BuildParamsFromStrings([]string)`, would be much easier to read/maintain.

qos/evm/endpoint.go
commoddity: I've added a todo in `state.go`'s `ValidateEndpoint` method:
```go
	// TODO_TECHDEBT(@commoddity): move the endpoint validation methods to the service state
	// and pass them the endpoint rather than passing the service state to each endpoint method.
```
Does this capture your intention?

qos/evm/endpoint.go
commoddity: Ditto above.

qos/evm/qos.go
commoddity: My assumption has been that we don't want to enforce that ALL services much be archival. For example, in the future we may want to allow running pruned EVM nodes in some services.

I know currently in Morse all Grove's chains are archival (F chains) but I don't think that this assumption should be baked into the code of PATH.

Tagging @Olshansk for input here as well.

qos/evm/state.go
commoddity: This is done.

See commit:
https://github.com/buildwithgrove/path/pull/194/commits/a4f04e376fbbe2c42bc4169ec9e4f8e900c0fd9d

qos/evm/state_archival.go
commoddity: This file has been refactored so this is no longer an issue.

qos/evm/state_archival.go
commoddity: Agreed; I've taken another pass over this file and I think it's much clearer now.

See commit:
https://github.com/buildwithgrove/path/pull/194/commits/a4f04e376fbbe2c42bc4169ec9e4f8e900c0fd9d

qos/evm/state_archival.go
commoddity: I've done so and moved the `archivalState` to be an argument rather than making it as method as this resolves the issue of splitting a struct method.

See commit:
https://github.com/buildwithgrove/path/pull/194/commits/cf70274f346fc5a97b0036219293e382661523cb

qos/evm/state_archival.go
commoddity: It's been largely refactored.

qos/jsonrpc/request.go
commoddity: OK, I've updated to remove `NewRequest` and instead create a ` jsonrpc.BuildArrayParamsFromStrings` helper method.

Please see commit: 
https://github.com/buildwithgrove/path/pull/194/commits/cf70274f346fc5a97b0036219293e382661523cb

qos/jsonrpc/request.go
```diff
@@ -23,30 +24,27 @@ type Request struct {
 	Params  Params  `json:"params,omitempty"`
 }
 
-// NewRequest constructs a new JSONRPC request from the given ID, method, and optional parameters.
+// BuildParamsFromStrings builds a Params object from an array of strings.
```
adshmh, line 27: Nice!

qos/jsonrpc/request.go
```diff
@@ -23,30 +24,27 @@ type Request struct {
...
-	if len(params) > 0 {
-		jsonParams, err := json.Marshal(params)
-		if err == nil {
-			request.Params = Params{rawMessage: jsonParams}
+// JSON-RPC array params must be passed in the order specified by the method.
+// Reference: https://www.jsonrpc.org/specification#parameter_structures
+//
+// TODO_FUTURE(@commoddity): other helper methods may be required to build
+// params for different JSON-RPC methods, eg. ["<string>", <bool>]
+func BuildArrayParamsFromStrings(params [2]string) (Params, error) {
```
adshmh, line 37: Not a blocker: 
Either:
-  Rename the method to emphasize it takes an array (and not a slice)
- Change the `params` type to `[]string`

qos/evm/check.go
```diff
@@ -51,3 +44,57 @@ func getEndpointCheck(endpointStore *EndpointStore, jsonrpcReq jsonrpc.Request)
...
+	}
+}
+
+// getArchivalCheckRequest returns a JSONRPC request to check the balance of:
+//   - the contract specified in `a.archivalCheckConfig.ContractAddress`
+//   - at the block number specified in `a.blockNumberHex`
+//
+// eg.
+// '{"jsonrpc":"2.0","id":1,"method":"eth_getBalance","params":["0x28C6c06298d514Db089934071355E5743bf21d60", "0xe71e1d"]}'
+func getArchivalCheckRequest(archivalState archivalState) (jsonrpc.Request, bool) {
```
adshmh, line 74: Great change moving the archival check builder function to `check.go`!

qos/evm/state.go
```diff
@@ -46,6 +46,9 @@ func (s *ServiceState) ValidateEndpoint(endpoint endpoint, endpointAddr protocol
 	s.serviceStateLock.RLock()
 	defer s.serviceStateLock.RUnlock()
 
+	// TODO_TECHDEBT(@commoddity): move the endpoint validation methods to the service state
```
adshmh, line 49: :+1:

adshmh (approved):
Thank you for addressing the review comments. Pre-approving.
Just one item to consider/update comment.

qos/jsonrpc/request.go
commoddity: Done.

Olshansk (approved):
@adshmh @commoddity Pushed a few changes and approving!

Note that I made some modifications based on how we define and instantinate `ServiceQoSConfigs`. It was based on my prior comments but I don't think I was clear enough with what I had in mind.

PTAL and review before taking next steps, but I think it's g2g!

Screenshot 2025-04-04 at 1 44 51 PM

commoddity:
> @adshmh @commoddity Pushed a few changes and approving!
> 
> Note that I made some modifications based on how we define and instantinate `ServiceQoSConfigs`. It was based on my prior comments but I don't think I was clear enough with what I had in mind.
> 
> PTAL and review before taking next steps, but I think it's g2g!
> 
> Screenshot 2025-04-04 at 1 44 51 PM

Changes all LGTM; I just pushed one minor nit PR to change the formatting for services with archival checks in `service_qos_config.go`. I'm fine with merging, WDYT? @Olshansk

adshmh:
> @adshmh @commoddity Pushed a few changes and approving!
> 
> Note that I made some modifications based on how we define and instantinate `ServiceQoSConfigs`. It was based on my prior comments but I don't think I was clear enough with what I had in mind.
> 
> PTAL and review before taking next steps, but I think it's g2g!
> 
LGTM.

Olshansk (approved):
💾 🚢 💾🚢💾 🚢 💾🚢💾 🚢 💾🚢💾 🚢 💾🚢💾 🚢 💾🚢

PR #241, opened 2025-04-29
[CI] speedup container image build

okdas:


Olshansk:
This was done in a different PR :)

PR #247, opened 2025-05-01
[Data Pipeline] Get trace ID from Envoy Proxy with fallback

commoddity:
## 🌿 Summary

Add support for getting trace/request ID from Envoy Proxy with fallback to UUID generation when not available.

### 🌱 Primary Changes:
- Added support for extracting trace ID from Envoy Proxy via `X-Request-Id` header
- Implemented fallback mechanism to generate UUID when trace ID is not available
- Updated trace ID generation logic to prefix synthetic IDs with "synthetic-"

## 🛠️ Type of change

## 🤯 Sanity Checklist


gateway/observation.go
```diff
@@ -1,6 +1,7 @@
 package gateway
```
Olshansk, line 1: Wdyt of having a single function to manage all the trace ID generation?
```go
// getOrGenerateTraceID determines the appropriate trace ID to use for a request.
// - If the request is nil, it's considered a synthetic request, and a synthetic trace ID is generated.
// - If the request is real, it attempts to extract the trace ID from the Envoy header (X-Request-Id).
//   If that header is missing, it falls back to generating a new UUID.
func getOrGenerateTraceID(httpReq *http.Request) string {
	if httpReq == nil {
		// Synthetic request: prepend "synthetic-" for observability/debugging.
		return fmt.Sprintf("synthetic-%s", uuid.New().String())
	}

	// Real request: try to extract from Envoy trace header first.
	if traceID := httpReq.Header.Get(httpHeaderEnvoyTraceID); traceID != "" {
		return traceID
	}

	// No trace ID present: generate a fallback one.
	return uuid.New().String()
}
```

Olshansk (changes requested):

gateway/observation.go
commoddity: > Wdyt of having a single function to manage all the trace ID generation?
> 
> ```go
> // getOrGenerateTraceID determines the appropriate trace ID to use for a request.
> // - If the request is nil, it's considered a synthetic request, and a synthetic trace ID is generated.
> // - If the request is real, it attempts to extract the trace ID from the Envoy header (X-Request-Id).
> //   If that header is missing, it falls back to generating a new UUID.
> func getOrGenerateTraceID(httpReq *http.Request) string {
> 	if httpReq == nil {
> 		// Synthetic request: prepend "synthetic-" for observability/debugging.
> 		return fmt.Sprintf("synthetic-%s", uuid.New().String())
> 	}
> 
> 	// Real request: try to extract from Envoy trace header first.
> 	if traceID := httpReq.Header.Get(httpHeaderEnvoyTraceID); traceID != "" {
> 		return traceID
> 	}
> 
> 	// No trace ID present: generate a fallback one.
> 	return uuid.New().String()
> }
> ```

IMO this makes the code less clear. The synthetic relays are a different enough case and I would prefer to keep this method single-concern: that is, generating the trace ID for an actual user request. Mixing the synthetic ID in there makes it confusing and is unnecessary given that all that synthetic IDs have only one possible source (`uuid.New()`).

Additionally we have a pending TODO to add the ability for the hydrator to use Delegated-mode requests which could make the `httpReq == nil` condition no longer apply here.

I would prefer to keep it as-is.

Olshansk: Fair enough and good point.

My ask would be to create a `getSyntheticTraceId` to make it explicit.

Stong preference but not a blocker.

Olshansk (approved):
LGTM but PTAL at this comment: https://github.com/buildwithgrove/path/pull/247/files#r2070901141

gateway/observation.go
commoddity: > Fair enough and good point.
> 
> My ask would be to create a `getSyntheticTraceId` to make it explicit.
> 
> Stong preference but not a blocker.

SG, I've implemented this and organized the file for more clarity.

PR #275, opened 2025-05-29
[Fix][Shannon] Increase cache TTL for apps to avoid too many requests to full node

commoddity:
## 🌿 Summary

Extend caching TTL for applications in Shannon to reduce the frequency of full node queries and improve performance.

### 🌱 Primary Changes:
- Increased app cache TTL from 30 seconds to 5 minutes to align with app unstaking behavior
- Set specific TTLs for apps and sessions rather than using shared default values
- Updated cache construction to use new TTL values for both expiration and cleanup

### 🍃 Secondary changes:
- Changed cache key formatting from slash-delimited (`prefix/key`) to hyphen-delimited (`prefix-key`)
- Left a TODO note to improve performance for fetching gateway-owned apps in the centralized mode implementation

## 💡 New TODOs 

- TODO_IMPROVE(@commoddity, @adshmh): This function currently loops through all apps owned by the gateway. Without a caching FullNode, this results in extremely slow behaviour. We should look into improving the efficiency of this lookup to get the list of apps owned by the gateway.  - protocol/shannon/mode_centralized.go

## 🛠️ Type of change

- [x] Bug fix

## 🤯 Sanity Checklist


protocol/shannon/fullnode_cache.go
```diff
@@ -58,9 +62,8 @@ type cachingFullNode struct {
 func NewCachingFullNode(lazyFullNode *lazyFullNode) *cachingFullNode {
 	return &cachingFullNode{
 		lazyFullNode: lazyFullNode,
-
-		appCache:     cache.New(defaultCacheTTL, defaultCacheCleanupInterval),
-		sessionCache: cache.New(defaultCacheTTL, defaultCacheCleanupInterval),
+		appCache:     cache.New(defaultAppCacheTTL, defaultAppCacheTTL*2),
```
adshmh, line 65: Could you please move the `*2` to the definition of `defaultAppCacheTTL`?

protocol/shannon/fullnode_cache.go
```diff
@@ -26,10 +26,21 @@ import (
...
+	// Cleanup interval is twice the TTL - standard practice for balancing
+	// timely cleanup with minimal processing overhead
+	defaultAppCacheCleanupInterval = defaultAppCacheTTL * 2
+
+	// A session is 10 blocks * 30 seconds, after that it is in grace period for 1 block (30 seconds).
+	// So it should be cached for - at most - 30 seconds.
+	defaultSessionCacheTTL = 30 * time.Second
+	// Cleanup interval is twice the TTL - standard practice for balancing
+	// timely cleanup with minimal processing overhead
+	defaultSessionCacheCleanupInterval = defaultSessionCacheTTL * 2
```
Olshansk, line 43:
```suggestion
	// Applications can be cached indefinitely.
	// They're invalidated only when they unstake.
	// TODO_MAINNET_MIGRATION: Ensure applications are invalidated during unstaking. Revisit these values after mainnet migration to ensure no race conditions.
	defaultAppCacheTTL = 5 * time.Minute

	// Cleanup interval is twice the TTL.
	// standard practice to balancing cleanup with minimal processing overhead
	defaultAppCacheCleanupInterval = defaultAppCacheTTL * 2

	// As of #275, on Beta TestNet.
	// - Blocks are 30 seconds
	// - A session is 50 blocks
	// - A grace period is 1 block (i.e. 30 seconds)
	// TODO_MAINNET_MIGRATION: Revisit these values after mainnet migration to ensure no race conditions.
	defaultSessionCacheTTL = 30 * time.Second
	
	// Cleanup interval is twice the TTL.
	// Standard practice for balancing cleanup with minimal processing overhead
	defaultSessionCacheCleanupInterval = defaultSessionCacheTTL * 2
```

protocol/shannon/fullnode_cache.go
```diff
@@ -177,8 +187,8 @@ func (cfn *cachingFullNode) GetAccountClient() *sdk.AccountClient {
 
 // createCacheKey creates a cache key for the given prefix and key.
 //
-//	eg. createCacheKey("app", "0x123") -> "app/0x123"
-//	eg. createCacheKey("session", "anvil:0x456") -> "session/anvil:0x456"
+//	eg. createCacheKey("app", "0x123") -> "app-0x123"
+//	eg. createCacheKey("session", "anvil:0x456") -> "session-anvil:0x456"
 func createCacheKey(prefix string, key string) string {
-	return fmt.Sprintf("%s/%s", prefix, key)
+	return fmt.Sprintf("%s-%s", prefix, key)
```
Olshansk, line 193: IMHO, we should be using either `:` or `/` for cache keys.

`-` is quite confusing.

Olshansk (changes requested):

PR #278, opened 2025-06-01
[Shannon][Migration][Protocol] Update caching full node to use SturdyC including early refresh

commoddity:
## 🌿 Summary

Enable automatic early-refresh caching for full node protocol using `SturdyC` to improve performance, reduce latency spikes, and prevent thundering herd issues.

Docs: https://github.com/viccon/sturdyc

### 🌱 Primary Changes:
- Replaced `go-cache` with `sturdyc` for both application and session caching.
- Implemented early refresh strategy with configurable TTL percentages for proactive background cache updates.
- Integrated `GetOrFetch` pattern for simplified cache access and built-in stampede protection.

### 🍃 Secondary changes:
- Removed legacy double-check locking mechanisms and manual cache mutex handling.
- Updated `go.mod` and `go.sum` to include `sturdyc` dependency.
- Added inline documentation describing the early refresh strategy and benefits.

## 🛠️ Type of change

## 🤯 Sanity Checklist


protocol/shannon/fullnode_cache.go
```diff
@@ -34,138 +33,136 @@ const (
...
 	// TODO_MAINNET_MIGRATION: Revisit these values after mainnet migration to ensure no race conditions.
 	defaultSessionCacheTTL = 30 * time.Second
-	
-	// Cleanup interval is twice the TTL.
-	// Standard practice for balancing cleanup with minimal processing overhead
-	defaultSessionCacheCleanupInterval = defaultSessionCacheTTL * 2
 
-	// Cache key prefixes to avoid collisions
+	// Early refresh configuration for apps
+	// Apps will start refreshing when they're between 4-4.5 minutes old
```
Olshansk, line 44: 1. #PUC what "refreshing" means and why we chose this number
2. Please add TODO_TECHDEBT(@olshansk) to revisit this.

protocol/shannon/fullnode_cache.go
```diff
@@ -34,138 +33,136 @@ const (
...
+	appMinRefreshDelay = 4 * time.Minute
+	appMaxRefreshDelay = 4*time.Minute + 30*time.Second
+
+	// Early refresh configuration for sessions
+	// Sessions will start refreshing when they're between 20-25 seconds old
```
Olshansk, line 49: Ditto about TODOs

protocol/shannon/fullnode_cache.go
```diff
@@ -34,138 +33,136 @@ const (
...
+	sessionMinRefreshDelay = 20 * time.Second
```
Olshansk, line 50: 1. Why such a small number?
2. Are we sure we want the bounds to be this small?

I'm thinking
min: 1 minute
max: 2 mins

protocol/shannon/fullnode_cache.go
```diff
@@ -34,138 +33,136 @@ const (
...
+	sessionMaxRefreshDelay = 25 * time.Second
+
+	// Retry base delay for exponential backoff on failed refreshes
+	retryBaseDelay = 100 * time.Millisecond
+
+	// Cache configuration
+	cacheCapacity      = 100_000
```
Olshansk, line 57: I don't know what these numbers are. Can you rename them to make it more semantic? Consider adding comments.

protocol/shannon/fullnode_cache.go
```diff
@@ -34,138 +33,136 @@ const (
...
-	cfn.sessionMutex.Lock()
-	defer cfn.sessionMutex.Unlock()
-
-	// Double-check cache after acquiring lock
-	if cachedSession, found := cfn.sessionCache.Get(cacheKey); found {
-		return cachedSession.(sessiontypes.Session), true
-	}
-
-	return sessiontypes.Session{}, false
+	// eg. "session:eth:pokt1up7zlytnmvlsuxzpzvlrta95347w322adsxslw"
```
Olshansk, line 155: Move the functionality/logic for sesison key generation to a separate file.

protocol/shannon/fullnode_cache.go
Olshansk: E.g. what do 'shards" refer to.

protocol/shannon/fullnode_cache.go
```diff
@@ -34,138 +33,136 @@ const (
...
 
 // cachingFullNode implements the FullNode interface by wrapping a LazyFullNode
-// and caching results to improve performance.
+// and caching results to improve performance with automatic refresh-ahead.
+//
+// Early Refresh Strategy:
+// Uses SturdyC's early refresh to prevent thundering herd and eliminate latency spikes.
+// Background refreshes happen before entries expire, so GetApp/GetSession never block.
+//
+//   - Apps: 5min TTL, refresh at 4-4.5min (80-90% of TTL)
```
Olshansk, line 76: Clarify that this is an example. Numbers will change. Add TODO on me.

Olshansk (changes requested):
This is great! I just want to capture some better comments we merge it in but otherwise let's go

protocol/shannon/fullnode_cache.go
commoddity: Because sessions last at most 30 seconds in the cache so the refresh delay must be smaller than that. 20-25 seconds is the range given which SturdyC uses to avoid a thundering herd problem of all cache refreshes happening at the same time if the TTL is similar.

protocol/shannon/fullnode_cache.go
commoddity: I could do a separate function but IMO a separate file is too much.

protocol/shannon/fullnode_cache.go
commoddity: Updated.

supplier_report_2025-06-02_21:16:40.json
```diff
@@ -0,0 +1,70 @@
+{
```
Olshansk, line 1: Is this intended to be ignored?

supplier_report_2025-06-02_21:02:28.json
```diff
@@ -0,0 +1,2191 @@
+{
```
Olshansk, line 1: Is this intended to be ignored?

Olshansk (approved):
I believ eyou want/need to remove a couple of files, but otherwise LGTM to me.

supplier_report_2025-06-02_21:16:40.json
commoddity: Oops, yes it is. The .gitignore line is in PR 268. Fixed now.

PR #292, opened 2025-06-10
[Shannon] Source apps from GetSession call to avoid possibility of apps being out of sync from sessions

commoddity:
## 🚨 TODO_IN_THIS_PR 

Do not merge until these TODOs are resolved:
- TODO_IN_THIS_PR: make this configurable in YAML with default value of 30 seconds - protocol/shannon/fullnode_cache.go

## 🌿 Summary

Refactor Shannon protocol to source apps from GetSession calls to ensure apps and sessions are always in sync, removing the need for separate app caching.

### 🌱 Primary Changes:
- Removed app caching layer and now source apps directly from session queries to prevent sync issues
- Updated centralized mode to fetch owned apps during initialization and use them for session queries
- Simplified caching strategy to only cache sessions with a shorter TTL (30s) to match protocol requirements

### 🍃 Secondary changes:
- Updated dependencies to latest poktroll and shannon-sdk versions
- Refactored lazyFullNode to LazyFullNode and improved naming conventions
- Removed unused app cache configuration and related constants

## 💡 New TODOs 

New TODOs introduced in this PR:
- TODO_MVP(@adshmh): Rename `LazyFullNode`: this struct does not perform any caching and should be named accordingly. - protocol/shannon/fullnode_lazy.go

## 🛠️ Type of change

Select one or more from the following:

- [ ] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## 🤯 Sanity Checklist

- [ ] I have updated the GitHub Issue 'assignees', 'reviewers', 'labels', 'project', 'iteration' and 'milestone'
- [ ] For docs, I have run 'make docusaurus_start'
- [ ] For code, I have run 'make test_all'
- [ ] For configurations, I have update the documentation
- [ ] I added TODOs where applicable

PR #298, opened 2025-06-11
[Review] Review of #292 which moved from GetApps to GetSessions

Olshansk:
## Summary

Refactor Shannon protocol initialization and simplify gateway mode session handling

### Primary Changes:
- Remove `ownedApps` parameter from `getShannonFullNode()` and move ownership logic to `NewProtocol()`
- Rename session-related functions from "permitted" to "valid" terminology throughout codebase
- Update function signatures to use `FullNode` interface instead of `*LazyFullNode` concrete type
- Consolidate centralized mode app initialization within protocol constructor

### Secondary changes:
- Add verbose logging with `@echo` statements to protobuf generation makefile targets
- Fix typo in signer setup error message (`"signe"` → `"signer"`)
- Compress documentation comments and improve code organization
- Update mock generation to use `github.com/golang/mock/gomock` instead of `go.uber.org/mock/gomock`
- Add TODO comments for session TTL alignment and critical restart requirements


protocol/shannon/protocol.go
```diff
@@ -187,16 +212,16 @@ func (p *Protocol) BuildRequestContextForEndpoint(
 		"endpoint_addr", selectedEndpointAddr,
 	)
 
-	permittedSessions, err := p.getGatewayModePermittedSessions(ctx, serviceID, httpReq)
+	validSessions, err := p.getValidGatewaySessions(ctx, serviceID, httpReq)
```
adshmh, line 215: I find the name `getValidGatewaySessions` confusing.:
- The `Valid` phrase could imply that `invalid` sessions may also be needed somewhere.
- `GatewaySession` part of the name can also connect Gateway to Session.
- The `GatewayModePermitted` part of the original phrase helped clarify these are sessions permitted under some gateway mode (it did need improvement).

adshmh (changes requested):
The main review comment I have is re naming: Valid vs Permitted and GatewayMode.
Would be great if we could find a name to convey some info about gateway modes and that they permit app/session to be used by the Gateway for sending relays.

protocol/shannon/config.go
```diff
@@ -141,6 +141,7 @@ func (c *GRPCConfig) hydrateDefaults() GRPCConfig {
 }
 
 // Session TTL should match the protocol's session length.
+// TODO_IMPROVE(@olshansk): Align this with the onchain session TTL (num blocks and num blocks per session).
```
commoddity, line 144: Ditto previous comment.

protocol/shannon/fullnode_account_fetcher.go
```diff
@@ -1,6 +1,9 @@
 package shannon
 
-// TODO_TECHDEBT(@commoddity): Refactor (remove?) this whole file based on the details in #291
+// TODO_TECHDEBT(@commoddity): Refactor (remove?) this whole file
```
commoddity, line 3: Confirming this file is removed in the subsequent refactor PRs, specifically as part 1, which is #295:
https://github.com/buildwithgrove/path/pull/295

PR #328, opened 2025-07-01
[Session] Fixing Session Rollovers using grace period

Olshansk:
## Summary

< ONE_LINE_SUMMARY>

### Primary Changes:

- < Change 1 >
- < Change 2 >

### Secondary Changes:

- < Change 1 >
- < Change 2 >

You can use the following as a helper for an LLM of your choice:

```bash
git --no-pager diff main  -- ':!*.pb.go' ':!*.pulsar.go' ':!*.json' ':!*.yaml' ':!*.yml' ':!*.gif' ':!*.md' | diff2html -s side --format json -i stdin -o stdout | pbcopy
```

## Issue

- Issue or PR: #{ISSUE_OR_PR_NUMBER}

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

### E2E Validation & Tests

- [ ] `make path_up`
- [ ] `make test_e2e_evm_shannon`
- [ ] `make test_e2e_evm_morse`

### Observability

- [ ] 1. `make path_up`
- [ ] 2. Run one of the following:
  - For `Shannon` with `anvil`: `make test_request__shannon_relay_util_100`
  - For `Morse` with `F00C`: `make test_request__morse_relay_util_100`
- [ ] 3. Visit [PATH Relay Grafana Dashboard](http://localhost:3003/d/relays/path-service-requests) to view results

## Sanity Checklist

- [ ] I have updated the GitHub Issue `assignees`, `reviewers`, `labels`, `project`, `iteration` and `milestone`
- [ ] For docs, I have run `make docusaurus_start`
- [ ] For code, I have run `make test_all`
- [ ] For configurations, I have updated the documentation
- [ ] I added `TODO`s where applicable

### DIFF
--- a/protocol/shannon/fullnode_cache.go
+++ b/protocol/shannon/fullnode_cache.go
@@ -10,6 +10,7 @@ import (
 	apptypes "github.com/pokt-network/poktroll/x/application/types"
 	servicetypes "github.com/pokt-network/poktroll/x/service/types"
 	sessiontypes "github.com/pokt-network/poktroll/x/session/types"
+	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
 	sdk "github.com/pokt-network/shannon-sdk"
 	"github.com/viccon/sturdyc"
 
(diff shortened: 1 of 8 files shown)


protocol/shannon/fullnode_cache.go
```diff
@@ -161,22 +215,111 @@ func (cfn *cachingFullNode) GetSession(
...
 	)
 }
 
-// getSessionCacheKey builds a unique cache key for session: <prefix>:<serviceID>:<appAddr>
-func getSessionCacheKey(serviceID protocol.ServiceID, appAddr string) string {
-	return fmt.Sprintf("%s:%s:%s", sessionCacheKeyPrefix, serviceID, appAddr)
+// GetSessionWithGracePeriod implements grace period aware session fetching with caching.
+// This method extends the standard GetSession with grace period logic while maintaining
+// the caching benefits for session data.
+func (cfn *cachingFullNode) GetSessionWithGracePeriod(
```
adshmh, line 241: `GetSessionWithGracePeriod` doesn't clarify it extends session validity.

`GetSessionWithExtendedValidity()` - shows sessions remain valid longer.
Or any other name that helps the reader understand the purpose.

protocol/shannon/fullnode_cache.go
```diff
@@ -161,22 +215,111 @@ func (cfn *cachingFullNode) GetSession(
...
 			)
 			return cfn.lazyFullNode.GetSession(fetchCtx, serviceID, appAddr)
 		},
 	)
 }
 
-// getSessionCacheKey builds a unique cache key for session: <prefix>:<serviceID>:<appAddr>
-func getSessionCacheKey(serviceID protocol.ServiceID, appAddr string) string {
-	return fmt.Sprintf("%s:%s:%s", sessionCacheKeyPrefix, serviceID, appAddr)
+// GetSessionWithGracePeriod implements grace period aware session fetching with caching.
```
adshmh, line 238: #PUC Current comment doesn't explain WHY this method exists.

Would be great to clarify:
- The distributed coordination problem it solves
- When it returns previous vs current session
- How it prevents request failures during rollovers

commoddity:
> Suggestions in addition to the review comments:
> 
> * Could we add some version of the following (assuming it is accurate) to the description and/or comments:
> 
> ```
> ## Problem
> PATH sends session data with every request. Suppliers don't switch sessions simultaneously, causing rollover failures.
> 
> ## Solution
> Continue using previous sessions during grace period transitions.
> 
> ## Result
> Prevents request failures during session rollovers.
> ```

I concur on these points. I've also reviewed the code and the solution seems reasonable but a brief overview of the problem/solution would have been helpful before starting the review to get a high-level picture.

> * Consider moving grace period logic to shannon-SDK:
>   
>   * Encapsulates protocol complexities in SDK (e.g. via Gateway component)
>   * PATH focuses on gateway logic.

@Olshansk I also wanted to add that I've created a draft PR as part of the Shannon SDK refactor work that I also believe will address the session mismatch error that is being tackled in this PR. 

You can see it here:
https://github.com/pokt-network/shannon-sdk/pull/41

commoddity (commented):
@Olshansk Just dropping this to let you know I've reviewed this PR as well; at a high level I don't see any issues with this approach. 

I do also agree with @adshmh 's comment regarding naming and detailed comments.

However, I did want to mention that I have also implemented a solution to the same problem in the Shannon SDK refactor work: https://github.com/pokt-network/shannon-sdk/pull/41

I haven't yet dug into both approaches enough to say which is "better" but I will say that what I've implemented does involve caching less different data types. Given the difficulties involved in caching multiple different data types over multiple different TTLs while trying to ensure that all the various data types remain "fresh" is a non-trivial task and for this reason I am biased towards an approach that caches as few different data types as possible.

For this reason, I think it would be worthwhile to investigate the approach I've tried in the Shannon SDK PR, which uses only one potentially-stale data type: the session itself. Then we can see if it satisfies the requirements of ensuring we do not use invalid session data.

CC @adshmh for thoughts on this as well.

Olshansk:
@commoddity

Appreciate the response.

My current thinking given my investigation (leaving details out):
- I'm in favor of your approach (long-term) after we figure out session rollovers and restore customer service
- I'm not confident this will solve the problem but WOULD LOVE to be proven wrong
- In this particular scenario, rather than reviewing code and evaluating tradeoffs: show me.

What I would do is:
- Build an image with your change
- Deploy it to a region (not the one I'm testing)
- Inspect if we do or don't have session rollovers there
- Add new metrics (if needed) to answer the question above

If we find that it solves session rollovers, we can throw this PR away altogether!

Olshansk:
Closing this in favor of #339 and #340 to test things separately.

PR #339, opened 2025-07-06
[Session] Introduce ExtendedSessions to RelayRequests

Olshansk:
> [!NOTE]
> Please see the `TODO` in `protocol/shannon/session.go`

## Primary Changes
- Implement `GetSessionWithExtendedValidity`
- Added 2 caches: `shared params` and `block height`
- Extracted session retrieval logic into a `session.go`

## Secondary changes
- Extract `full_node` interface into its own file
- Modify some timeout configs
- Misc QoL & code health improvements


local/path/values.tmpl.yaml
```diff
@@ -19,16 +19,18 @@ guard:
   # 1. Each service must have a unique service ID.
   # 2. Each service can have a list of aliases that can be used to route requests to the service.
   services:
```
commoddity, line 21: Adding both `serviceId` and `aliases` with the same values is not needed as the `HTTPRoute` resources are created by making a new route for each of:
`serviceId + (aliases...)`

In my experience having duplicate entries for the same string values can also causes k8s issues in non-Tilt environments.

I suggest we do something like this:
```yaml
  services:
    - serviceId: eth
      aliases:
        - ethereum
    - serviceId: anvil
    - serviceId: pocket
    - serviceId: poly
      aliases:
        - polygon
```

This way we have examples for how aliases may be used (`eth/ethereum` and `poly/polygon`) while removing any redundant duplication of the same strings. IMO this makes the intended function of aliases clearer to users of PATH.

protocol/shannon/full_node.go
```diff
@@ -0,0 +1,75 @@
...
+
+// FullNode defines the set of capabilities the Shannon protocol integration needs
+// from a fullnode for sending relays.
+//
+// A properly initialized fullNode struct can:
+// 1. Return the onchain apps matching a service ID.
+// 2. Fetch a session for a (service,app) combination.
+// 3. Validate a relay response.
+// 4. Etc...
+type FullNode interface {
```
commoddity, line 23: Good call to move this to its own file. 👍

protocol/shannon/fullnode_cache.go
```diff
@@ -136,11 +160,43 @@ func NewCachingFullNode(
...
+		sturdyc.WithEarlyRefreshes(
+			sharedParamsMinRefreshDelay,
+			sharedParamsMaxRefreshDelay,
+			sharedParamsCacheTTL,
+			retryBaseDelay,
+		),
+	)
+
+	// Block height cache
+	blockHeightMinRefreshDelay, blockHeightMaxRefreshDelay := getCacheDelays(blockHeightCacheTTL)
```
commoddity, line 179: Given the consistent interval on which block height changes we may want to consider making the blockHeight delay a const rather than calculating based on the same percentages as the other caches, ie. make it early refresh at a consistent 15 seconds. Given we only cache one entry for block height I don't think we need to worry about thundering herd issues, which is main reason for staggering the early cache refresh for the session cache.

protocol/shannon/fullnode_cache.go
```diff
@@ -207,3 +374,39 @@ func (cfn *cachingFullNode) GetAccountClient() *sdk.AccountClient {
...
+				cfn.logger.Error().Err(fetchErr).Msg("Failed to fetch shared params from full node")
+			}
+			return params, fetchErr
+		},
+	)
+
+	return params, err
+}
+
+// GetCurrentBlockHeight: cached block height with 20sec TTL and early refresh.
```
commoddity, line 396:
```suggestion
// GetCurrentBlockHeight: cached block height with short TTL and early refresh.
```

protocol/shannon/mode_delegated.go
```diff
@@ -31,46 +31,67 @@ func (p *Protocol) getDelegatedGatewayModeActiveSession(
...
+				// selectedSessions = append(selectedSessions, sessionPreviousExtended)
+				// TODO_IN_THIS_PR: Revert this change.
+				// We are experimenting by forcing it to always use the previous session.
+				selectedSessions = []sessiontypes.Session{sessionPreviousExtended}
+				logger.Info().Msg("EXPERIMENT: Overriding the latest session with the previous session for the app.")
+			}
+		}
+	}
 
+	// Select the first session in the list.
```
commoddity, line 74: Ditto

commoddity (changes requested):
Looks great to me. Just a few minor comments, no real blockers but not approving yet since there are `TODO_IN_THIS_PR` comments remaining.

protocol/morse/protocol.go
```diff
@@ -132,7 +132,7 @@ func (p *Protocol) AvailableEndpoints(
 // BuildRequestContextForEndpoint builds a new request context for a given service ID and endpoint address.
 // Implements the gateway.Protocol interface.
 func (p *Protocol) BuildRequestContextForEndpoint(
-	_ context.Context,
+	ctx context.Context,
```
adshmh, line 135: NIT: ctx is not used in the method. Using `_` as name makes the code more readable by clarifying this.

PR #341, opened 2025-07-07
[Session] Improve relay response metrics and timeout propogation

Olshansk:
- Main process: better management of kill signal
- Config fix: s/MaxRequestBodySize/MaxRequestHeaderBytes
- Full Node: Extract the full node interface into its own file
- Historgram: New historgram to track response sizes
- Backup timeout for relay requests
- Code health: A bit of misc cleanup across the board


protocol/shannon/relay.go
```diff
@@ -44,11 +46,21 @@ func sendHttpRelay(
 	//
 	// Best practice in Go is to use a custom HTTP client Transport.
 	// See: https://vishnubharathi.codes/blog/know-when-to-break-up-with-go-http-defaultclient/
-	relayHTTPResponse, err := http.DefaultClient.Do(relayHTTPRequest)
+	client := &http.Client{
+		Timeout: timeout,
```
adshmh, line 50: This looks redundant: the HTTP request has a timeout set on it:
https://github.com/buildwithgrove/path/blob/69ea48321885e09095984302c2dbea5f8a30b20a/protocol/shannon/relay.go#L31-L32

protocol/shannon/observation.go
```diff
@@ -166,6 +168,12 @@ func buildEndpointObservation(
 	observation.Supplier = endpoint.supplier
 	observation.EndpointUrl = endpoint.url
 
+	// Add endpoint response details if not nil (i.e. success)
+	if endpointResponse != nil {
+		*observation.EndpointResponseHttpStatusCode = int32(endpointResponse.HTTPStatusCode)
```
adshmh, line 173: I believe this is the HTTP status code returned by the backend service, relayed to PATH via RelayMiner and [deserialized via the SDK](https://github.com/pokt-network/shannon-sdk/blob/e527d4172770cba5606678fdd381b5481e124788/types/response.go#L51)
The HTTP status code from the call to RelayMiner is ignored in `sendHttpRelay`.

Just making sure this is what we intend to capture here. If so, I suggest a different name for the field to clarify this point.

adshmh (changes requested):
Just a few nits, main one on naming the HTTP status code field to clarify its origin.

metrics/protocol/shannon/metrics.go
```diff
@@ -395,6 +395,8 @@ func processEndpointLatency(
 				"service_id":      serviceID,
 				"endpoint_domain": endpointTLDPlusOne,
 				"success":         fmt.Sprintf("%t", success),
+				"http_status":     fmt.Sprintf("%d", endpointObs.GetEndpointResponseHttpStatusCode()),
+				"response_size":   fmt.Sprintf("%d", endpointObs.GetEndpointResponseHttpPayloadSize()),
```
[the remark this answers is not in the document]
adshmh: In addition, the cardinality from this is going to break Grafana.
Looks like some bucketing logic was intended (based on label name `request_size_bucket`) but is missing from the PR.

Olshansk:
@adshmh Thank you for the review here.

Tended to everything you suggested, found some other issues along the way and deployed it directly to see if we find any changes: https://github.com/buildwithgrove/infrastructure/commit/207081116931c5078cede83cf451eebafcde8bfe

adshmh (changes requested):
Marking as `Request changes` since the `response_size` metric label cardinality could impact production.

adshmh (approved):
LGTM

PR #361, opened 2025-07-17
[QoS][Cosmos] feat: handle empty responses from the supplier for cosmos

commoddity:
## 🌿 Summary

Handle and standardize empty responses from the Cosmos supplier to improve compatibility across REST and JSON-RPC protocols.

### 🌱 Primary Changes:
- Introduced handling for empty endpoint responses by adding a specific error response when response data is empty
- Implemented `getEmptyJSONRPCErrResponse` to return structured error responses for empty payloads
- Differentiated logic based on whether the incoming request is JSON-RPC or REST-like
- Updated response ID assignment to use `restLikeResponseID` for non-JSON-RPC requests with empty IDs

### 🍃 Secondary changes:
- Replaced deprecated and commented-out ID logic with simplified, protocol-aware decision making
- Improved consistency in error response creation by passing `isJSONRPC` to unmarshaller and error helpers
- Renamed variables and comments to reflect REST-like handling (instead of JSON-RPC specific language) for clarity

## 🛠️ Type of change

### DIFF
--- a/.github/workflows/run-e2e-tests.yml
+++ b/.github/workflows/run-e2e-tests.yml
@@ -83,6 +83,7 @@ jobs:
 
       - name: update Shannon E2E config from secrets
         env:
+          SHANNON_GATEWAY_ADDRESS: ${{ secrets.SHANNON_GATEWAY_ADDRESS }}
           SHANNON_GATEWAY_PRIVATE_KEY: ${{ secrets.SHANNON_GATEWAY_PRIVATE_KEY }}
           SHANNON_OWNED_APPS_PRIVATE_KEYS: ${{ secrets.SHANNON_OWNED_APPS_PRIVATE_KEYS}}
         run: ./e2e/scripts/update_shannon_config_from_secrets.sh


qos/cosmos/response_generic.go
```diff
@@ -17,6 +17,12 @@ const (
 	// errMsgUnmarshaling is the generic message returned to the user if the endpoint returns a malformed response.
 	errMsgUnmarshaling = "the response returned by the endpoint is not a valid JSON-RPC response"
 
+	// errCodeEmptyResponse is the error code returned to the user if the endpoint returns an empty response.
```
adshmh, line 20: This should be -32000 to be consistent with other qos package.

adshmh (changes requested):
Thanks, just needs a response code update.

qos/cosmos/response.go
```diff
@@ -69,7 +52,7 @@ func unmarshalResponse(
 	if err := json.Unmarshal(data, &jsonrpcResponse); err != nil {
 		// The response raw payload could not be unmarshalled as a JSON-RPC response.
 		// Treat it as a REST response and use the generic unmarshaller.
-		return responseUnmarshallerGeneric(logger, jsonrpcResponse, data)
+		return responseUnmarshallerGeneric(logger, jsonrpcResponse, data, isJSONRPC)
```
Olshansk, line 55: Do we even need `isJSONRPC`?

Feels like this should be `false`, the one below should be `true` and then we don't need to pass in the extra variable?

Olshansk (commented):
@adshmh I don't want to block this PR, but can you take a holistic look?

It's getting a bit messy and feels like we can potentially just remove the `isJSONRPC` flag altogether. It's just a hunch.

It's fine to do in a followup, but do spend 5-10 minutes thinking about it.

PR #363, opened 2025-07-17
[QoS][Cosmos] Use HTTP request's path to determine request type

adshmh:
## Summary

Use HTTP request's path to determine request type

### Primary Changes:

- Use HTTP request's path to determine request type

## Issue

- Issue or PR: #{ISSUE_OR_PR_NUMBER}

## Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

### E2E Validation & Tests

- [ ] `make path_up`
- [ ] `make test_e2e_evm_shannon`

### Observability

- [ ] 1. `make path_up`
- [ ] 2. Run the following E2E test: `make test_request__shannon_relay_util_100`
- [ ] 3. View results in LocalNet's [PATH Relay Grafana Dashboard](http://localhost:3003/d/relays/path-service-requests)

## Sanity Checklist

- [ ] I have updated the GitHub Issue `assignees`, `reviewers`, `labels`, `project`, `iteration` and `milestone`
- [ ] For docs, I have run `make docusaurus_start`
- [ ] For code, I have run `make test_all`
- [ ] For configurations, I have updated the documentation
- [ ] I added `TODO`s where applicable

### DIFF
--- a/qos/cosmos/qos.go
+++ b/qos/cosmos/qos.go
@@ -5,7 +5,6 @@ import (
 	"net/http"
 
 	"github.com/pokt-network/poktroll/pkg/polylog"
-	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
 
 	"github.com/buildwithgrove/path/gateway"
 	"github.com/buildwithgrove/path/metrics/devtools"
@@ -56,17 +55,11 @@ func NewQoSInstance(logger polylog.Logger, config CosmosSDKServiceQoSConfig) *Qo
 	}
 
 	requestValidator := &requestValidator{
-		logger:       logger,
-		serviceID:    serviceId,
-		chainID:      cosmosSDKChainID,
-		serviceState: serviceState,
-
-		// TODO_UPNEXT(@adshmh): Make supported RPC types configurable.
-		supportedAPIs: map[sharedtypes.RPCType]struct{}{
-			sharedtypes.RPCType_REST:      {},
-			sharedtypes.RPCType_COMET_BFT: {},
-			sharedtypes.RPCType_JSON_RPC:  {},
-		},
+		logger:        logger,
+		serviceID:     serviceId,
+		chainID:       cosmosSDKChainID,
+		serviceState:  serviceState,
+		supportedAPIs: config.getSupportedAPIs(),
 	}
 
 	return &QoS{
--- a/qos/cosmos/service_qos_config.go
+++ b/qos/cosmos/service_qos_config.go
@@ -26,19 +26,19 @@ type CosmosSDKServiceQoSConfig interface {
 	ServiceQoSConfig // Using locally defined interface to avoid circular dependency
 	getCosmosSDKChainID() string
 	getSyncAllowance() uint64
-	getRPCTypes() map[sharedtypes.RPCType]struct{}
+	getSupportedAPIs() map[sharedtypes.RPCType]struct{}
 }
 
 // NewCosmosSDKServiceQoSConfig creates a new CosmosSDK service configuration.
 func NewCosmosSDKServiceQoSConfig(
 	serviceID protocol.ServiceID,
 	cosmosSDKChainID string,
-	rpcTypes map[sharedtypes.RPCType]struct{},
+	supportedAPIs map[sharedtypes.RPCType]struct{},
 ) CosmosSDKServiceQoSConfig {
 	return cosmosSDKServiceQoSConfig{
 		serviceID:        serviceID,
 		cosmosSDKChainID: cosmosSDKChainID,
-		rpcTypes:         rpcTypes,
+		supportedAPIs:    supportedAPIs,
 	}
 }
 
@@ -49,7 +49,7 @@ type cosmosSDKServiceQoSConfig struct {
 	serviceID        protocol.ServiceID
 	cosmosSDKChainID string
 	syncAllowance    uint64
-	rpcTypes         map[sharedtypes.RPCType]struct{}
+	supportedAPIs    map[sharedtypes.RPCType]struct{}
 }
 
 // GetServiceID returns the ID of the service.
@@ -79,14 +79,16 @@ func (c cosmosSDKServiceQoSConfig) getSyncAllowance() uint64 {
 	return c.syncAllowance
 }
 
-// getRPCTypes returns the RPC types supported by the service.
+// getSupportedAPIs returns the RPC types supported by the service.
 // For example, XRPLEVM supports the following RPC types:
 //   - JSON_RPC
 //   - REST
 //   - COMET_BFT
 //   - WEBSOCKET (does not currently have a QoS quality check system in PATH)
 //
-// This is used to determine the appropriate QoS endpoint checks to run
-func (c cosmosSDKServiceQoSConfig) getRPCTypes() map[sharedtypes.RPCType]struct{} {
-	return c.rpcTypes
+// This is used:
+// 1. to validate the request and whether this service supports the request's RPC type
+// 2. to determine the appropriate syntheticQoS endpoint checks to run
+func (c cosmosSDKServiceQoSConfig) getSupportedAPIs() map[sharedtypes.RPCType]struct{} {
+	return c.supportedAPIs
 }
--- a/qos/cosmos/service_state.go
+++ b/qos/cosmos/service_state.go
@@ -62,13 +62,13 @@ func (ss *serviceState) GetRequiredQualityChecks(endpointAddr protocol.EndpointA
 	endpoint := ss.endpointStore.endpoints[endpointAddr]
 
 	// Get the RPC types supported by the CosmosSDK service.
-	rpcTypes := ss.serviceQoSConfig.getRPCTypes()
+	supportedAPIs := ss.serviceQoSConfig.getSupportedAPIs()
 
 	// List of all checks required for the endpoint.
 	var checks []gateway.RequestQoSContext
 
 	// If the service supports CometBFT, add the CometBFT endpoint checks.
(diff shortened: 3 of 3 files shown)


qos/cosmos/request_validator.go
```diff
@@ -1,217 +1,239 @@
...
-	// Empty body POST requests are treated as REST
-	// Example: POST /some/endpoint with no payload
-	if len(body) == 0 {
-		return crv.createRESTRequestContext(req, body), true
+	// Priority 2: CometBFT endpoints (can be REST or JSON-RPC based on HTTP method)
+	// Examples: /health, /status, /genesis, /validators, /block, etc.
+	if isCometBftRpc(path) {
+		if method == http.MethodPost {
+			// POST to CometBFT endpoints can contain JSON-RPC payloads
+			return sharedtypes.RPCType_JSONRPC
```
commoddity, line 83: @adshmh I think that the approach here regarding request validation is generally in the right direction. 

My concern with this PR is that the RPC-type terminology may be getting overloaded. 

What I mean is that in the context of Cosmos SDK sending headers to the Relay Miner, the RPC-types have very specific meanings, as in they map to specific endpoints on the Relay Miner backend.

1. `RPCType_COMET_BFT` - CometBFT (port 26657)
2. `RPCType_REST` - Cosmos SDK (port 1317)
3. `RPCType_JSONRPC` - EVM JSON-RPC (port 8545)

So in the context of request validation all of your changes make perfect sense: a request can either be JSON-RPC or it can be REST-like.

But then when getting the `Rpc-Type` header that is sent to the Relay Miner backend, the RPC type meanings are slightly different since there are 3 possible options. This process happens in `getRPCTypeHeaders` in `qos/cosmos/rpctype_headers.go`

- `EVM JSON-RPC` can be detected by checking for the EVM method prefixes (eg `eth_`, `web3_`, etc.)
- `Cosmos SDK` can be detected by the URL paths (eg `/cosmos/`)
- `CometBFT` is detected by either URL paths (if REST-like - eg `/status/`) or methods (if JSON-RPC - eg `"method": "status"`)

So there are two different contexts where this enum is used and its meaning is slightly different in both.

Just want to make sure this part is clear. Overall the PR is in the right direction but I wonder if there's a way to make these differences clear so that the `RPCType` enum is not overloading because that will inevitable confuse us later... or now. 😅

commoddity (changes requested):
Great PR. Must appreciate the clarity and cleanup it brings.

My main thoughts are in this comment:

qos/cosmos/qos.go
```diff
@@ -54,17 +55,24 @@ func NewQoSInstance(logger polylog.Logger, config CosmosSDKServiceQoSConfig) *Qo
 		endpointStore:    store,
 	}
 
-	cosmosSDKRequestValidator := &cosmosSDKRequestValidator{
+	requestValidator := &requestValidator{
 		logger:       logger,
 		serviceID:    serviceId,
 		chainID:      cosmosSDKChainID,
 		serviceState: serviceState,
```
commoddity, line 62: The request validator does not contain the required `supportedAPIs` map, which causes requests to fail in local testing.

I've opened a small PR to fix this and to align the language of what was previously called `rpcTypes` to the terminology used in this PR (`supportedAPIs`): https://github.com/buildwithgrove/path/pull/365

qos/cosmos/request_validator.go
```diff
@@ -1,217 +1,118 @@
...
-	return &requestContext{
-		logger:               crv.logger,
-		httpReq:              *req,
-		chainID:              crv.chainID,
-		serviceID:            crv.serviceID,
-		requestPayloadLength: payloadLength,
-		serviceState:         crv.serviceState,
-		requestOrigin:        qosobservations.RequestOrigin_REQUEST_ORIGIN_ORGANIC,
+	// Stage 2: POST requests - check for JSONRPC payload
+	if strings.Contains(string(httpRequestBody), "jsonrpc") {
```
commoddity, line 73: Nice! This is a great check that does not require unmarshaling.

qos/cosmos/request_validator.go
```diff
@@ -1,217 +1,118 @@
...
-// - chainID: The CosmosSDK chain identifier for which the request was intended
-// - err: The error that occurred during HTTP body reading
-//
-// Returns:
-// - qosobservations.Observations: A structured observation containing details about the HTTP read failure
 func createHTTPBodyReadFailureObservation(
 	serviceID protocol.ServiceID,
 	chainID string,
 	err error,
+	jsonrpcResponse jsonrpc.Response,
```
commoddity, line 105: NIT:
```suggestion
	jsonrpcResponse jsonrpc.Response,
	err error,
```

qos/cosmos/request_validator_jsonrpc.go
```diff
@@ -0,0 +1,318 @@
...
+	method := string(jsonrpcReq.Method)
+	rpcType := detectJSONRPCServiceType(method)
+
+	// Hydrate the logger with data extracted from the request.
+	rv.logger = logger.With(
+		"detected_rpc_type", rpcType.String(),
+		"jsonrpc_method", method,
+	)
+
+	// Check if this RPC type is supported by the service
```
commoddity, line 53: #PUC that the `supportedAPIs` map is configured in `config/service_qos_config.go`.

qos/cosmos/request_validator_rest.go
```diff
@@ -0,0 +1,246 @@
...
+
+	// Determine the specific RPC type based on path patterns - use existing function
+	rpcType := determineRESTRPCType(httpRequestPath)
+
+	logger = logger.With(
+		"detected_rpc_type", rpcType.String(),
+		"request_path", httpRequestPath,
+	)
+
+	// Check if this RPC type is supported by the service
```
commoddity, line 41: Ditto previous comment re: this map being configured in `config/service_qos_config.go.`

qos/cosmos/request_validator_rest.go
```diff
@@ -0,0 +1,246 @@
...
+	if _, supported := rv.supportedAPIs[rpcType]; !supported {
+		logger.Warn().Msg("Request uses unsupported RPC type")
+		// TODO_TECHDEBT(@adshmh): Review error context creation for REST requests.
```
commoddity, line 44: WDYT about using `-1` as the ID field for error responses for CometBFT requests? 

This seems to be the standard they use, eg. all responses to a `/status` REST endpoint return a JSON-RPC response with ID = `-1`.

qos/cosmos/request_validator_rest.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+}
+
+// TODO_TECHDEBT(@adshmh): Review the expected user experience on protocol errors in REST requests.
+func buildRESTProtocolErrorResponse() func(logger polylog.Logger) gateway.HTTPResponse {
+	return func(logger polylog.Logger) gateway.HTTPResponse {
+		// For REST requests, we return a JSON-RPC error response with null ID
+		// TODO_TECHDEBT(@adshmh): Consider returning proper REST error response format
+		errorResp := jsonrpc.NewErrResponseInternalErr(
+			jsonrpc.ID{}, // use null as ID.
```
commoddity, line 175: Ditto here. Perhaps we should determine the error response based on RPC type since CometBFT always returns JSON-RPC responses even for REST-like requests.

For CosmosSDK I agree that perhaps it should be REST error response however.

qos/cosmos/response_health.go
```diff
@@ -1,90 +1,74 @@
...
+	"github.com/buildwithgrove/path/gateway"
 	qosobservations "github.com/buildwithgrove/path/observation/qos"
+	"github.com/buildwithgrove/path/qos"
 	"github.com/buildwithgrove/path/qos/jsonrpc"
+	"github.com/pokt-network/poktroll/pkg/polylog"
 )
 
-// responseToHealth provides the functionality required from a response by a requestContext instance.
-var _ response = responseToHealth{}
+// responseValidatorHealth implements jsonrpcResponseValidator for /health endpoint
```
commoddity, line 11: Not a blocker and this should be done in a subsequent PR but just making a note that we should likely clarify that certain methods are CometBFT vs Cosmos SDK vs EVM JSON-RP (eg. `/health` is specifically CometBFT).

qos/cosmos/response_health.go
```diff
@@ -1,90 +1,74 @@
...
+
+	return &responseHealth{
+		logger:              logger,
+		userJSONRPCResponse: jsonrpcResponse,
+		healthy:             true,
+	}
 }
 
-// responseToHealth captures a CometBFT-based blockchain's /health endpoint response.
+// responseHealth captures a Cosmos JSONRPC /health endpoint response
```
commoddity, line 41:
```suggestion
// responseHealth captures a CometBFT /health endpoint response
```

qos/cosmos/rpctype_cosmos.go
```diff
@@ -1,35 +1,20 @@
 package cosmos
 
-import "strings"
+import (
+	sharedtypes "github.com/pokt-network/poktroll/x/shared/types"
+)
 
 // ------------------------------------------------------------------------------------------------
-// Cosmos SDK RPC Type Detection
+// REST API Type Detection - Consolidated Logic
```
commoddity, line 8: Perhaps rename this file to `qos/cosmos/rpctype_rest.go`?

commoddity (changes requested):
Looking really great. I like the direction this has taken quite a bit. It cleans up the logic and makes things much more clearly separated and maintainable.

PTAL @ https://github.com/buildwithgrove/path/pull/365 as it uses the proper configuration for supported APIS from the `config/service_qos_config.go` file and fixes some naming issues.

PR #391, opened 2025-08-04
[Gateway][Config] Add Fallback URL configuration and Gateway package handling for fallback requests

commoddity:
## 🌿 Summary

Add fallback URL configuration and Shannon protocol package handling for fallback requests when no protocol endpoints are available.

### 🌱 Primary Changes:
- Added `service_fallback_endpoint_urls` configuration in gateway config schema with validation
- Implemented fallback endpoint handling in Shannon protocol with separate request flow
- Added protocol observation flag for fallback endpoints
- Refactored relay request handling into separate protocol and fallback paths

### 🍃 Secondary changes:
- Added unit tests for fallback URL validation
- Updated documentation placeholder for fallback URL configuration
- Improved code organization with helper methods for shared request logic

## 🛠️ Type of change

### DIFF
--- a/protocol/shannon/config.go
+++ b/protocol/shannon/config.go
@@ -73,15 +73,11 @@ type (
 	// - Criteria to decide whether the "fallback" URL should be used at all.
 	// - Criteria to decide the order in which a Shannon endpoint vs. a fallback URL should be used.
 	// - Support "weighted" distribution to Shannon endpoints vs. "fallback" URLs.
-	// - Remove the `serviceFallbackConfig` struct: update and use `ServiceFallback` as the canonical struct everywhere.
 	//
 	// ServiceFallback is a configuration struct for specifying fallback endpoints for a service.
 	ServiceFallback struct {
 		ServiceID    protocol.ServiceID `yaml:"service_id"`
 		FallbackURLs []string           `yaml:"fallback_urls"`
-
-		// TODO_TECHDEBT(@adshmh): Explicitly define and implement what happens if this is false.
-		//
 		// If true, all traffic will be sent to the fallback endpoints for the service,
 		// regardless of the health of the protocol endpoints.
 		SendAllTraffic bool `yaml:"send_all_traffic"`


config/fallback.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+// FallbackURLs is a map of service IDs to fallback URLs.
+// It is unmarshaled from the YAML field `fallback_urls`.
+//
+// Example YAML:
+//
+//	fallback_urls:
+//	  eth: https://eth.rpc.grove.city/v1/1a2b3c4d
+type FallbackURLs map[protocol.ServiceID]string
```
adshmh, line 17: I recommend avoiding new types unless they server a very specific purpose: Doesn't seem like Fallback URLs should be a self managing entity.

gateway/request.go
```diff
@@ -13,6 +14,11 @@ type HTTPRequestParser interface {
 	// GetQoSService returns the qos for the service matching an HTTP request.
 	GetQoSService(context.Context, *http.Request) (protocol.ServiceID, QoSService, error)
 
+	// GetFallbackURL returns the fallback URL to use in case
+	// no endpoints are available for the requested service.
+	// Returns the fallback URL and a boolean indicating whether a fallback URL is configured.
+	GetFallbackURL(context.Context, *http.Request) (*url.URL, bool)
```
adshmh, line 20: Similar general note re placement of fallback specification/validation/handling mechanisms: protocol/shannon package seems a better fit. I also recommend to avoid promoting the request parser to handle anything besides request parsing.

proto/path/gateway.proto
```diff
@@ -59,6 +61,12 @@ message GatewayObservations {
 
   // parallel_request_observations tracks the outcome of parallel requests within a batch.
   optional GatewayParallelRequestObservations gateway_parallel_request_observations = 8;
+
+  // fallback_used indicates whether a fallback URL was used for the request.
+  bool fallback_used = 9;
```
adshmh, line 66: Similarly for other fallback* fields: shannon.proto (or a new file under proto/path/protocol/) seems like a  better fit.

protocol/shannon/endpoint.go
```diff
@@ -31,6 +31,16 @@ type endpoint struct {
 	session sessiontypes.Session
 }
 
+// fallbackSupplier is a const value used to identify fallback endpoints.
+// Fallback endpoints do not existing on the Shannon protocol and so do not have a supplier address.
```
adshmh, line 35: `s/existing/exist/`

protocol/shannon/context.go
```diff
@@ -200,74 +197,105 @@ func buildHeaders(payload protocol.Payload) map[string]string {
...
+	// Validate and process the response
+	response, err := rc.validateAndProcessResponse(hydratedLogger, httpRelayResponseBz)
+	if err != nil {
+		return protocol.Response{}, err
+	}
+
+	// Deserialize the response
+	deserializedResponse, err := rc.deserializeRelayResponse(response)
+	if err != nil {
+		return protocol.Response{}, err
```
adshmh, line 243: Same here re endpoint URL in case of error.

adshmh (changes requested):
Great improvement, just a few minor items to address.

Olshansk (commented):
Sending an initial (stale) review from yesterday since it seems outdated.

Going to restart shrotly.

protocol/shannon/config.go
```diff
@@ -60,11 +61,25 @@ type (
 		SessionTTL time.Duration `yaml:"session_ttl"`
 	}
 
+	ServiceFallback struct {
```
adshmh, line 64: #PUC on the purpose of this struct.

protocol/shannon/config.go
```diff
@@ -60,11 +61,25 @@ type (
...
+		ServiceID      protocol.ServiceID `yaml:"service_id"`
+		SendAllTraffic bool               `yaml:"send_all_traffic"`
+		FallbackURLs   []string           `yaml:"fallback_urls"`
+	}
+
+	// ServiceFallbackConfig holds the fallback configuration for a service,
+	// including the endpoints and whether to send all traffic to fallback.
+	ServiceFallbackConfig struct {
```
adshmh, line 72: Do we have to make this public? It makes the code confusing since there is also a `ServiceFallback` struct.

protocol/shannon/protocol.go
```diff
@@ -296,6 +318,16 @@ func (p *Protocol) getSessionsUniqueEndpoints(
 		serviceID, len(activeSessions),
 	)
 
+	// If the service is configured to send all traffic to fallback endpoints,
+	// return only the fallback endpoints and skip the rest of the logic.
+	if p.getSendAllTrafficToFallback(serviceID) {
```
adshmh, line 323: NIT: the method name / how we access this configuration item could be improved.

protocol/shannon/config.go
commoddity: Yes, you're correct. This can be private and is better located in the `protocol.go` file. I've made this change.

Olshansk (commented):
Partial review (to expedite things).

protocol/shannon/protocol.go
```diff
@@ -55,6 +56,20 @@ type Protocol struct {
...
+	// The fallback endpoints are used when no endpoints are available for
+	// the requested service. For example, if all protocol endpoints are sanctioned,
+	// the fallback endpoints will be used to populate the list of endpoints.
+	// Each service can have a SendAllTraffic flag to control traffic routing behavior.
+	fallbackEndpoints map[protocol.ServiceID]serviceFallbackConfig
+}
+
+// serviceFallbackConfig holds the fallback configuration for a service,
+// including the endpoints and whether to send all traffic to fallback.
+type serviceFallbackConfig struct {
```
Olshansk, line 70: WHy not reuse the structure in `protocol/shannon/config.go`?

protocol/shannon/http_client.go
```diff
@@ -124,9 +123,9 @@ func (h *httpClientWithDebugMetrics) SendHTTPRelay(
 	ctx context.Context,
```
Olshansk, line 123: #PUC of godoc of what it retusn

protocol/shannon/protocol.go
```diff
@@ -55,6 +56,20 @@ type Protocol struct {
 
 	// HTTP client used for sending relay requests to endpoints while also capturing & publishing various debug metrics.
 	httpClient *httpClientWithDebugMetrics
+
+	// fallbackEndpoints contains the fallback endpoint configurations for the protocol.
+	// The fallback endpoints are used when no endpoints are available for
+	// the requested service. For example, if all protocol endpoints are sanctioned,
+	// the fallback endpoints will be used to populate the list of endpoints.
+	// Each service can have a SendAllTraffic flag to control traffic routing behavior.
```
Olshansk, line 64:
```suggestion
	// fallbackEndpoints contains the fallback endpoint configurations for the protocol.
	//
	// The fallback endpoints are used when no endpoints are available for the requested service
	// from the onchain protocol.
	// For example, if all protocol endpoints are sanctioned,  the fallback endpoints will be used to
	// populate the list of endpoints.
	//
	// Each service can have a SendAllTraffic flag to control traffic routing behavior.
```

Olshansk, line 64: #PUC related to SendAllTraffic for more details.

config/config_test.go
```diff
@@ -162,6 +172,102 @@ logger_config:
...
+			      host_port: "shannon-testnet-grove-grpc.beta.poktroll.com:443"
+			  gateway_config:
+			    gateway_mode: "centralized"
+			    gateway_address: "pokt1up7zlytnmvlsuxzpzvlrta95347w322adsxslw"
+			    gateway_private_key_hex: "40af4e7e1b311c76a573610fe115cd2adf1eeade709cd77ca31ad4472509d388"
+			    owned_apps_private_keys_hex:
+			      - "40af4e7e1b311c76a573610fe115cd2adf1eeade709cd77ca31ad4472509d388"
+			    service_fallback:
+			      - service_id: ""
+			        send_all_traffic: false
```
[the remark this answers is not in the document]
commoddity: The traffic would be evenly distributed by QoS among all valid fallback URLs in the same way that QoS handles any other endpoints (QoS has no knowledge of protocol-level data like whether an endpoint is a fallback or not).

protocol/shannon/protocol.go
```diff
@@ -296,7 +325,21 @@ func (p *Protocol) getSessionsUniqueEndpoints(
...
 
+	// Get fallback configuration for the service ID.
+	// Fallback configuration will be used in the following scenarios:
+	// 	 1. The service is configured to send all traffic to fallback endpoints.
+	// 	 2. All service endpoints are sanctioned and the service is configured with at least one fallback endpoint.
+	fallbackEndpoints, sendAllTrafficToFallback := p.getServiceFallbackConfig(serviceID)
+
+	// If the service is configured to send all traffic to fallback endpoints,
+	// return only the fallback endpoints and skip the rest of the logic.
+	if sendAllTrafficToFallback && len(fallbackEndpoints) > 0 {
```
Olshansk, line 336: IMO this SHOULD NOT be part of `getSessionUniqueEndpoints`.

We're hiding/embedding logic.

I'd prefer this return an empty list/array and the caller will then retrieve fallbackEndpoints.

cc @adshmh for another POV.

config/config_test.go
```diff
@@ -162,6 +172,102 @@ logger_config:
...
+			      host_port: "shannon-testnet-grove-grpc.beta.poktroll.com:443"
+			  gateway_config:
+			    gateway_mode: "centralized"
+			    gateway_address: "pokt1up7zlytnmvlsuxzpzvlrta95347w322adsxslw"
+			    gateway_private_key_hex: "40af4e7e1b311c76a573610fe115cd2adf1eeade709cd77ca31ad4472509d388"
+			    owned_apps_private_keys_hex:
+			      - "40af4e7e1b311c76a573610fe115cd2adf1eeade709cd77ca31ad4472509d388"
+			    service_fallback:
+			      - service_id: eth
+			        send_all_traffic: false
```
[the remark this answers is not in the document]
commoddity: No I don't think we should do this. This adds a lot of complexity and is out of scope for this PR.

protocol/shannon/protocol.go
Olshansk: You can create a `getUniqueEndpoints` that wraps this logic.

protocol/shannon/endpoint.go
```diff
@@ -31,6 +31,16 @@ type endpoint struct {
 	session sessiontypes.Session
 }
 
+// fallbackSupplier is a const value used to identify fallback endpoints.
+// Fallback endpoints do not exist on the Shannon protocol and so do not have a supplier address.
+// Instead, they are identified by the fallbackSupplier const value.
+const fallbackSupplier = "fallback"
```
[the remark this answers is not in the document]
commoddity: IMO `grove` does not make sense here. The `fallback` const is essentially just a placeholder. `fallback` accurately describes the functionality of the endpoint, `grove` does not.

docusaurus/docs/develop/configs/2_gateway_config.md
```diff
@@ -123,12 +133,42 @@ shannon_config:
...
+    - "https://poly.rpc.grove.city/v1/9x8y7z6w"
+```
+
+| Field          | Type     | Description                                                                                                                                                            |
+| -------------- | -------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `<service_id>` | string[] | Array of fallback endpoint URLs for the specified service ID. Each URL must be a valid HTTP/HTTPS endpoint. **Service ID must be a valid onchain Shannon service ID.** |
+
+**Key Features:**
+- **Automatic failover**: Gateway automatically switches to fallback endpoints when protocol endpoints are unavailable
+- **Multiple endpoints per service**: Configure multiple fallback URLs for redundancy and load distribution
```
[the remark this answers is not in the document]
commoddity: Fallback endpoint are selected by QoS in the same way as any other endpoints: the QoS package will pick a random valid endpoint. The "fallback" concept is encapsulated entirely in the `protocol` package so the `qos` package has no knowledge of whether an endpoint is fallback or not.

I'll update the docs to indicate this.

docusaurus/docs/develop/configs/2_gateway_config.md
```diff
@@ -123,12 +133,42 @@ shannon_config:
...
+  eth:
+    - "https://eth.rpc.grove.city/v1/1a2b3c4d"
+    - "https://eth.rpc.grove.city/v1/5e6f7a8b"
+  poly:
+    - "https://poly.rpc.grove.city/v1/9x8y7z6w"
+```
+
+| Field          | Type     | Description                                                                                                                                                            |
+| -------------- | -------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `<service_id>` | string[] | Array of fallback endpoint URLs for the specified service ID. Each URL must be a valid HTTP/HTTPS endpoint. **Service ID must be a valid onchain Shannon service ID.** |
```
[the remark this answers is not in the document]
commoddity: I've updated it now.

proto/path/protocol/shannon.proto
```diff
@@ -200,6 +200,9 @@ message ShannonEndpointObservation {
 
   // HTTP Response payload size
   optional int64 endpoint_backend_service_http_response_payload_size = 15;
+
+  // Whether the endpoint is a fallback endpoint
```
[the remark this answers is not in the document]
commoddity: Ditto previous comment re: this naming.

Olshansk (changes requested):
Continuing partial review

protocol/shannon/config.go
```diff
@@ -60,11 +61,25 @@ type (
...
+		SendAllTraffic bool
```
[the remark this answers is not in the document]
commoddity: 1. Comment is outdated; one struct has private to the protocol package and has been relocated.
2. I do not think it should be. Perhaps in a future TODO but I believe traffic splitting is far out of scope for this initial PR.

protocol/shannon/protocol.go
commoddity: They are not the same:

`protocol/shannon/protocol.go`
```go
type serviceFallbackConfig struct {
	SendAllTraffic bool
	Endpoints      map[protocol.EndpointAddr]endpoint
}
```

`protocol/shannon/config.go`
```go
ServiceFallback struct {
		ServiceID    protocol.ServiceID `yaml:"service_id"`
		FallbackURLs []string           `yaml:"fallback_urls"`
		// If true, all traffic will be sent to the fallback endpoints for the service,
		// regardless of the health of the protocol endpoints.
		SendAllTraffic bool `yaml:"send_all_traffic"`
	}
)```

File: protocol/shannon/protocol.go (hunk 13 above)
Comment (commoddity) [reply protocol/shannon/protocol.go:336]: This is fine, and I agree it makes it cleaner. I've refactored to separate session endpoint logic.

File: config/config.schema.yaml [hunk 20]
```diff
@@ -110,6 +110,30 @@ properties:
             items:
               type: string
               pattern: "^[0-9a-fA-F]{64}$"
+          service_fallback:
```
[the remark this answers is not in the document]
adshmh: Fair point re potentially confusing the reader, but could we just add the TODO_CONSIDERATION for later evaluation? It does not rename anything yet, just leaves a note for future reconsideration of the name and/or usage.

protocol/shannon/protocol.go
adshmh: Agreed that making the logic more explicit and visible is a good change here.

config/config.schema.yaml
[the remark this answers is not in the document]
Olshansk: My goal is to signal to develoeprs (us) that this will extend/be changed in the future.

I'm not asking for change sin this PR.

Olshansk (approved):
Approving contingent on two things:

1. @adshmh does another round of reviews and adds TODO_TECHDEBT items for future itertions/improvements/considerations
2. This was validated locally by both @adshmh and @commoddity

protocol/shannon/protocol.go
commoddity: IMO protocol is the correct place for it. It's used directly only in the protocol related code and is a property of the `Protocol` struct. It is not a config level struct.

config/config.schema.yaml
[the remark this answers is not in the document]
commoddity: OK that makes sense.

protocol/shannon/config.go
```diff
@@ -65,6 +66,25 @@ type (
...
+	// - Criteria to decide the order in which a Shannon endpoint vs. a fallback URL should be used.
+	// - Support "weighted" distribution to Shannon endpoints vs. "fallback" URLs.
+	// - Remove the `serviceFallbackConfig` struct: update and use `ServiceFallback` as the canonical struct everywhere.
+	//
+	// ServiceFallback is a configuration struct for specifying fallback endpoints for a service.
+	ServiceFallback struct {
+		ServiceID    protocol.ServiceID `yaml:"service_id"`
+		FallbackURLs []string           `yaml:"fallback_urls"`
+
+		// TODO_TECHDEBT(@adshmh): Explicitly define and implement what happens if this is false.
```
commoddity, line 83: This is already implemented: fallback endpoints are selected by the `qos` package in the same way as any other endpoints.

PR #396, opened 2025-08-08
[Protocol][Shannon] Shannon: add a parallel fallback request after 1 second

adshmh:
## Summary

Shannon: add a parallel fallback request after 1 second

### Primary Changes:

- Shannon: add a parallel fallback request after 1 second

## Issue

Success rate drop due to endpoint slowness.

- Issue or PR: #{ISSUE_OR_PR_NUMBER}

## Type of change

Select one or more from the following:

- [x] New feature, functionality or library
- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

### E2E Validation & Tests

- [ ] `make path_up`
- [ ] `make test_e2e_evm_shannon`

### Observability

- [ ] 1. `make path_up`
- [ ] 2. Run the following E2E test: `make test_request__shannon_relay_util_100`
- [ ] 3. View results in LocalNet's [PATH Relay Grafana Dashboard](http://localhost:3003/d/relays/path-service-requests)

## Sanity Checklist

- [ ] I have updated the GitHub Issue `assignees`, `reviewers`, `labels`, `project`, `iteration` and `milestone`
- [ ] For docs, I have run `make docusaurus_start`
- [ ] For code, I have run `make test_all`
- [ ] For configurations, I have updated the documentation
- [ ] I added `TODO`s where applicable

PR #397, opened 2025-08-08
[QoS] feat: add sessionRolloverState and session rollover awareness to full node

commoddity:
## 🌿 Summary

Add session rollover awareness to full node implementation to improve reliability during session transitions.

### 🌱 Primary Changes:
- Implement `sessionRolloverState` to track session transition periods
- Add background monitoring for session rollover periods via `startSessionRolloverMonitoring`
- Expose `IsInSessionRollover()` method to check current rollover status

### 🍃 Secondary changes:
- Convert log messages to structured JSON format in `cmd/main.go`
- Add passthrough implementation for `IsInSessionRollover()` in caching full node
- Add detailed logging for session and block height changes

## 💡 New TODOs 

- TODO_NEXT(@adshmh): Incorporate this with changes in PR #396 to only use parallel fallback request when in session rollover period Reference: https://github.com/buildwithgrove/path/pull/396  IsInSessionRollover returns true if the system is currently in a session rollover period.  A session rollover period is a critical time window that occurs around session transitions and can cause reliability issues for relay operations. The rollover period is defined as: - 1 block before the session end height - Plus a configurable grace period after the session end (default: 10 blocks)  This method enables the gateway to implement adaptive retry strategies during rollover periods  The monitoring is performed automatically in the background and this method provides a thread-safe way to check the current rollover status. - protocol/shannon/full_node.go
- TODO_TECHDEBT(@commoddity): Make this configurable via config file Grace period after session end where rollover issues may occur - protocol/shannon/fullnode_session_rollover.go

## 🛠️ Type of change

- [ ] Bug fix
- [ ] Code health or cleanup
- [ ] Documentation
- [ ] Other (specify)

## 🤯 Sanity Checklist

- [ ] I have updated the GitHub Issue 'assignees', 'reviewers', 'labels', 'project', 'iteration' and 'milestone'
- [ ] For docs, I have run 'make docusaurus_start'
- [ ] For code, I have run 'make test_all'
- [ ] For configurations, I have update the documentation
- [ ] I added TODOs where applicable

### DIFF
--- a/protocol/shannon/fullnode_lazy.go
+++ b/protocol/shannon/fullnode_lazy.go
@@ -42,10 +42,15 @@ type LazyFullNode struct {
 	blockClient   *sdk.BlockClient
 	accountClient *sdk.AccountClient
 	sharedClient  *sdk.SharedClient
+
+	// Session rollover monitoring state
+	rolloverState *sessionRolloverState
 }
 
 // NewLazyFullNode builds and returns a LazyFullNode using the provided configuration.
 func NewLazyFullNode(logger polylog.Logger, config FullNodeConfig) (*LazyFullNode, error) {
+	logger = logger.With("component", "fullnode_lazy")
+
 	// Hydrate defaults for all config sections
 	config.hydrateDefaults()
 
@@ -81,7 +86,11 @@ func NewLazyFullNode(logger polylog.Logger, config FullNodeConfig) (*LazyFullNod
 		blockClient:   blockClient,
 		accountClient: accountClient,
 		sharedClient:  sharedClient,
+		rolloverState: &sessionRolloverState{},
 	}
+
+	// Start session rollover monitoring in background
+	fullNode.startSessionRolloverMonitoring()
 
 	return fullNode, nil
 }
@@ -122,6 +131,9 @@ func (lfn *LazyFullNode) GetSession(
 				serviceID, appAddr, err,
 			)
 	}
+
+	// Update session end height for rollover monitoring
+	lfn.updateSessionEndHeight(*session)
 
 	return *session, nil
 }
@@ -237,6 +249,13 @@ func (lfn *LazyFullNode) GetSessionWithExtendedValidity(
 
 	// Return the previous session
 	return *prevSession, nil
+}
+
+// IsInSessionRollover returns true if we're currently in a session rollover period.
+// A session rollover period is problematic for relay operations and typically lasts
+// about 5 minutes (10 blocks) after a session ends.
+func (lfn *LazyFullNode) IsInSessionRollover() bool {
+	return lfn.getSessionRolloverState()
 }
 
 // serviceRequestPayload:
--- a/protocol/shannon/fullnode_session_rollover.go
+++ b/protocol/shannon/fullnode_session_rollover.go
@@ -1,0 +1,159 @@
+package shannon
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	sessiontypes "github.com/pokt-network/poktroll/x/session/types"
+)
+
+const (
+	// TODO_TECHDEBT(@commoddity): Make this configurable via config file
+	// Grace period after session end where rollover issues may occur
+	sessionRolloverGracePeriodBlocks = 10
+
+	// How often to check for block height updates
+	blockCheckInterval = 15 * time.Second
+)
+
+// sessionRolloverState tracks session rollover status for the LazyFullNode.
+//
+// Session rollovers are critical periods that occur around session transitions
+// where relay operations may experience higher failure rates due to:
+// - Session synchronization issues between gateways and suppliers
+// - Temporary endpoint unavailability during session changes
+// - Inconsistent session state across the network
+//
+// This state is maintained automatically by background monitoring that:
+// 1. Periodically fetches the current block height
+// 2. Updates when new sessions are retrieved via GetSession()
+// 3. Calculates rollover status based on the rollover window
+//
+// The rollover window spans from 1 block before session end through
+// sessionRolloverGracePeriodBlocks after session end. This provides
+// early warning and extended monitoring of potentially problematic periods.
+//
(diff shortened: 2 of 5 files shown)


protocol/shannon/context.go
```diff
@@ -124,6 +113,40 @@ func (rc *requestContext) HandleServiceRequest(payload protocol.Payload) (protoc
 	return relayResponse, err
 }
 
+// executeRelayRequest determines and executes the appropriate relay strategy based on:
+//  1. Endpoint type (fallback vs protocol endpoint)
+//  2. Network conditions (session rollover periods)
+func (rc *requestContext) executeRelayRequest(payload protocol.Payload) (protocol.Response, error) {
+	logger := rc.logger.With("method", "executeRelayRequest")
```
adshmh, line 120: Please add a `service_id` label.

protocol/shannon/fullnode_lazy.go
```diff
@@ -123,6 +132,9 @@ func (lfn *LazyFullNode) GetSession(
 			)
 	}
 
+	// Update session rollover boundaries for rollover monitoring
+	lfn.updateSessionValues(*session)
```
adshmh, line 136: `updateSessionRollverBoundaries` seems like a better name (or any other name that indicates what the method actually does)

protocol/shannon/fullnode_session_rollover.go
```diff
@@ -0,0 +1,167 @@
...
+	"sync"
+	"time"
+
+	sessiontypes "github.com/pokt-network/poktroll/x/session/types"
+)
+
+const (
+	// TODO_IN_THIS_PR(@commoddity): Make this configurable via config file
+	// Grace period after session end where rollover issues may occur
+	sessionRolloverGracePeriodBlocks = 10
```
adshmh, line 14: I believe we are at 60 blocks per session, setting this to a more aggressive initial value of 5 seems to make more sense considering our goal here.

protocol/shannon/fullnode_session_rollover.go
```diff
@@ -0,0 +1,167 @@
...
+	lfn.logger.Info().
+		Dur("check_interval", blockCheckInterval).
+		Int("grace_period_blocks", sessionRolloverGracePeriodBlocks).
+		Msg("Starting session rollover monitoring")
+
+	go lfn.monitorLoop()
+}
+
+// monitorLoop continuously checks block height to detect session rollovers
+func (lfn *LazyFullNode) monitorLoop() {
```
adshmh, line 64: Please update the method name to indicate what it does.

protocol/shannon/fullnode_session_rollover.go
```diff
@@ -0,0 +1,167 @@
...
+func (lfn *LazyFullNode) startSessionRolloverMonitoring() {
+	lfn.logger.Info().
+		Dur("check_interval", blockCheckInterval).
+		Int("grace_period_blocks", sessionRolloverGracePeriodBlocks).
+		Msg("Starting session rollover monitoring")
+
+	go lfn.monitorLoop()
+}
+
+// monitorLoop continuously checks block height to detect session rollovers
```
adshmh, line 63: NIT: this is done periodically, not continuously.

protocol/shannon/fullnode_session_rollover.go
```diff
@@ -0,0 +1,167 @@
...
+	if lfn.rolloverState.sessionRolloverEnd == 0 || currentBlockHeight > lfn.rolloverState.sessionRolloverEnd {
+		newRolloverStart := sessionEndHeight - 1
+		newRolloverEnd := sessionEndHeight + sessionRolloverGracePeriodBlocks
+
+		oldRolloverStart := lfn.rolloverState.sessionRolloverStart
+
+		// Update rollover boundaries and recalculate rollover status
+		lfn.rolloverState.sessionRolloverStart = newRolloverStart
+		lfn.rolloverState.sessionRolloverEnd = newRolloverEnd
+		lfn.rolloverState.isInSessionRollover = lfn.calculateRolloverStatus()
```
adshmh, line 99: Please merge this with the call at the bottom, to remove the need for `else` clause.

protocol/shannon/fullnode_session_rollover.go
```diff
@@ -0,0 +1,167 @@
...
+	// Only update rollover boundaries if current rollover period has ended
+	// or if we don't have rollover boundaries set yet
+	if lfn.rolloverState.sessionRolloverEnd == 0 || currentBlockHeight > lfn.rolloverState.sessionRolloverEnd {
+		newRolloverStart := sessionEndHeight - 1
+		newRolloverEnd := sessionEndHeight + sessionRolloverGracePeriodBlocks
+
+		oldRolloverStart := lfn.rolloverState.sessionRolloverStart
+
+		// Update rollover boundaries and recalculate rollover status
+		lfn.rolloverState.sessionRolloverStart = newRolloverStart
```
adshmh, line 97: Moving all these changes into a method of the rolloverState struct would clarify things quite a bit.

protocol/shannon/fullnode_session_rollover.go
commoddity: Generally our observed errors during session rollover happen during a 5 minute period, so my estimate was:
` 10 blocks* 30 seconds = 5 minutes` 
as a starting point. Do you think this is incorrect?

commoddity: > Trying to minimize the use of fallback endpoints, hence the suggestion for a shorter window during which we use fallback.

My concern is that:
1. a shorter window will not cover the full period where we experience elevated error rates
2. this PR turns off ALL fallback outside this window

Therefore we have the risk of seeing error rates during session rollover go back up.

Going from using fallback 100% of the time to 1/6th of the time is already a huge reduction so IMO it's better to err on the side of caution and not risk error rates rising due to turning off fallback during a period where error rates may be higher than usual.

PR #424, opened 2025-08-28
[Websockets][QoS] Add websockets QoS checks with separate protocol-level endpoint sanction stores for websockets and HTTP

commoddity:
## 🌿 Summary

Add WebSocket QoS checks with separate protocol-level endpoint sanction stores for WebSocket and HTTP RPC types.

### 🌱 Primary Changes:
- Add WebSocket-specific endpoint sanction store to maintain separate sanctions for WebSocket and HTTP RPC types
- Implement WebSocket connection checks in the endpoint hydrator with separate intervals for HTTP and WebSocket checks
- Extend protocol interface with WebSocket-specific methods for endpoint availability, observation application, and connection checking
- Add WebSocket support configuration to EVM service QoS configs (enabled for base and bsc chains)

### 🍃 Secondary changes:
- Refactor endpoint hydrator into separate HTTP and WebSocket check implementations for better maintainability
- Update metrics reporting to handle protocol-level disqualified endpoints per RPC type
- Add WebSocket connection check support to EVM, Cosmos, and NoOp QoS implementations
- Extend WebSocket observation handling with connection and message-level observations for better monitoring

## 💡 New TODOs 

- TODO_TECHDEBT(@commoddity): Currently websocket QoS only performs a protocol-level check which determines if an endpoint connection request is successful or not.  In the future, we may want to add QoS-level checks that take into account specific endpoint responses to apply websocket-related filtering at the QoS level. - gateway/qos.go
- TODO_TECHDEBT(@adshmh,@commoddity,@olshansk): JSON_RPC RPC type should more correctly be called HTTP when used in this context. Add an HTTP RPC-type to the enum in poktroll and update this map when it is done.  sanctionedEndpointsStores tracks sanctioned endpoints per RPC type currently only JSON_RPC (stand-in for HTTP) and WEBSOCKET are supported - protocol/shannon/protocol.go

## 🛠️ Type of change

### DIFF
--- a/gateway/hydrator_websocket.go
+++ b/gateway/hydrator_websocket.go
@@ -60,7 +60,7 @@ func (eph *EndpointHydrator) performWebSocketChecks(serviceID protocol.ServiceID
 	// Passing a nil as the HTTP request, because we assume the hydrator uses "Centralized Operation Mode".
 	// TODO_TECHDEBT(@adshmh): support specifying the app(s) used for sending/signing synthetic relay requests by the hydrator.
 	// TODO_FUTURE(@adshmh): consider publishing observations if endpoint lookup fails.
-	availableEndpoints, _, err := eph.AvailableEndpoints(context.TODO(), serviceID, nil)
+	availableEndpoints, _, err := eph.AvailableWebsocketEndpoints(context.TODO(), serviceID, nil)
 	if err != nil || len(availableEndpoints) == 0 {
 		// No session found or no endpoints available for service: skip.
 		logger.Warn().Msg("no session found or no endpoints available for service when running WebSocket hydrator checks.")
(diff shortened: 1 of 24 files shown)


gateway/hydrator_websocket.go
```diff
@@ -0,0 +1,129 @@
...
+	serviceID protocol.ServiceID,
+	endpointAddr protocol.EndpointAddr,
+) error {
+	// Create a context with a short timeout for the WebSocket connection check
+	// This ensures we don't wait too long for unresponsive endpoints
+	checkTimeout := 10 * time.Second
+	ctx, cancel := context.WithTimeout(context.Background(), checkTimeout)
+	defer cancel()
+
+	obs := eph.Protocol.CheckWebsocketConnection(ctx, serviceID, endpointAddr)
```
adshmh, line 123: Please add a TODO: this is an internal detail of protocol (similar to e.g. fetching sessions).
It should be encapsulated and handled automatically inside protocol (e.g. via a goroutine started at the time of protocol instance initialization), as there is no input required from any other components.

This is different from endpoint quality checks, where QoS needs to provide the payload to send to the endpoint.
Protocol can perform regular WS checks against in-session endpoints and adjust the list of available endpoints for WS requests accordingly.

adshmh (approved):
Thank you for the PR.
Please make sure to add TODOs before merging.
Top item to consider/address later is exposing internal details of protocol package for WS checks (rather than encapsulating them similar to how the rest of the code base has no knowledge of sessions).

PR #429, opened 2025-09-03
[Fix][QoS] Ensure backend status codes result in sanctions

commoddity:
## 🌿 Summary

Ensure backend service HTTP status codes result in proper session sanctions when they return non-2xx responses.

### 🌱 Primary Changes:
- Added HTTP status code validation for backend service responses in Shannon protocol
- Implemented session sanctions for backend services returning non-2xx status codes
- Created reusable CheckHTTPStatusCode function for HTTP status validation

### 🍃 Secondary changes:
- Refactored HTTP status code checking into a shared utility function
- Enhanced error messages to distinguish between relay endpoint and backend service errors
- Updated sanction logic to handle backend service HTTP errors specifically

## 🛠️ Type of change

- [x] Bug fix


Olshansk (approved):
LGTM! 

Pushed some minor edits and cleanup and approving.

Will deploy to PROD shortly.

Got claude looking into other test improvements in parallel but will add those in a separate PR if it's good.

PR #444, opened 2025-09-10
[Bug] fix issue with extracting TLD from endpoint address

commoddity:
## 🌿 Summary

Fix an issue with extracting TLD from endpoint addresses by implementing robust domain extraction with proper fallback handling.

Example error:
```json
{
  "level": "error",
  "method": "PublishMetricsEVM",
  "error": "publicsuffix: cannot derive eTLD+1 for domain \"pkp\"",
  "message": "SHOULD NEVER HAPPEN: Cannot get endpoint domain from endpoint address: pokt1htpka5l37qef6swtpqn7qwalv6jezu5rre5anl-https://pkp-og.relayminer.shannon-mainnet.eu.nodefleet.net"
}
```

### 🌱 Primary Changes:
- Replace simple dash-splitting with first-dash detection to handle URLs containing dashes
- Implement comprehensive domain extraction using golang.org/x/net/publicsuffix for proper TLD+1 detection
- Add fallback logic for edge cases like IP addresses, localhost, and internal domains
- Remove dependency on metricshttp package to avoid import cycles

### 🍃 Secondary changes:
- Add error handling for empty URL parts and malformed URLs
- Include helper functions for identifying localhost and private/internal domains
- Copy domain extraction logic from path/metrics/protocol/shannon/domain.go to maintain functionality

## 🛠️ Type of change

- [ ] Bug fix

### DIFF
--- a/config/service_qos_config.go
+++ b/config/service_qos_config.go
@@ -49,350 +49,639 @@ var shannonServices = []ServiceQoSConfig{
 	// *** EVM Services (Archival) ***
 
 	// Arbitrum One
-	evm.NewEVMServiceQoSConfig("arb-one", "0xa4b1", evm.NewEVMArchivalCheckConfig(
-		// https://arbiscan.io/address/0xb38e8c17e38363af6ebdcb3dae12e0243582891d
-		"0xb38e8c17e38363af6ebdcb3dae12e0243582891d",
-		// Contract start block
-		3_057_700,
-	)),
+	evm.NewEVMServiceQoSConfig(
+		"arb-one", "0xa4b1",
+		evm.NewEVMArchivalCheckConfig(
+			// https://arbiscan.io/address/0xb38e8c17e38363af6ebdcb3dae12e0243582891d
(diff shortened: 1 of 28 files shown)


protocol/endpoint.go
```diff
@@ -55,29 +57,122 @@ func (e EndpointAddr) String() string {
...
-	// Take everything before the first dash as the address
-	return endpointAddrParts[0], nil
+	// Extract the supplier address part (everything before the first dash)
+	return e.String()[:dashIndex], nil
+}
+
+// extractDomainOrHost extracts the effective TLD+1 from a URL.
+// It falls back to a reasonable domain extraction for localhost, IP addresses, and other non-standard hosts.
+// This is a copy of the logic from path/metrics/protocol/shannon/domain.go to avoid import cycles.
+func extractDomainOrHost(rawURL string) (string, error) {
```
adshmh, line 101: This does not look like a protocol responsibility.
Can we move this to the `metrics/protocol` or `metrics/protocol/shannon` package? (and reuse/extend any supporting functionality that already exists in that package).

PR #448, opened 2025-09-15
[Bug][E2E] Adds several minor bug fixes and E2E support for eth, base and bsc

commoddity:
## 🌿 Summary

Adds Ethereum WebSocket support to E2E tests and bug fixes for endpoint domain extraction for fallback endpoints.

### 🌱 Primary Changes:
- Enable Ethereum WebSocket testing in E2E CI workflow by adding eth service to test matrix
- Add WebSocket RPC type support to service QoS configuration for proper metrics tracking
- Implement WebSocket-specific endpoint URL handling for fallback and public endpoints in observation logic
- Extend WebSocket test validation to support multiple blockchain services beyond just xrplevm

### 🍃 Secondary changes:
- Improve logging throughout WebSocket connection handling and metrics processing
- Refactor error handling comments and documentation in EVM interpreter for clarity
- Fix endpoint URL extraction in WebSocket connection establishment failure observations
- Standardize logger context method naming across metrics processing functions

### DIFF
--- a/e2e/config/services_shannon.yaml
+++ b/e2e/config/services_shannon.yaml
@@ -33,6 +33,7 @@ services:
     service_type: "evm"
     alias: "arbitrum-one"
     archival: true
+    websockets: false
     service_params:
       # https://arbiscan.io/address/0xb38e8c17e38363af6ebdcb3dae12e0243582891d
       contract_address: "0xb38e8c17e38363af6ebdcb3dae12e0243582891d"
@@ -47,6 +48,7 @@ services:
     service_type: "evm"
     alias: "arbitrum-sepolia-testnet"
     archival: true
+    websockets: false
     service_params:
       # https://sepolia.arbiscan.io/address/0x22b65d0b9b59af4d3ed59f18b9ad53f5f4908b54
       contract_address: "0x22b65d0b9b59af4d3ed59f18b9ad53f5f4908b54"
@@ -60,6 +62,7 @@ services:
     evm_chain_id: "0xa86a"
     service_type: "evm"
     archival: true
+    websockets: false
     service_params:
       # https://avascan.info/blockchain/c/address/0x9f8c163cBA728e99993ABe7495F06c0A3c8Ac8b9
       contract_address: "0x9f8c163cBA728e99993ABe7495F06c0A3c8Ac8b9"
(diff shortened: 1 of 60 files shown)


e2e/config/services_shannon.yaml
```diff
@@ -100,6 +101,7 @@ services:
     alias: "base-sepolia-testnetnet"
     service_type: "evm"
     archival: true
+    websockets: true
```
Olshansk, line 104: Can you set `websockets: false` on the other ones?

It'll make it easier to toggle/read/understand for readers.

e2e/service_evm_test.go
```diff
@@ -86,7 +86,6 @@ func getEVMTestMethodsForWebSocket() []string {
 		eth_getBalance,
 		eth_getBlockByNumber,
 		eth_getTransactionCount,
-		eth_getTransactionReceipt,
```
Olshansk, line 89: Screenshot 2025-09-16 at 5 08 18 PM

We still reference it in other parts of the codebase.

Why did we remove it here?

Olshansk (changes requested):
No hard blockers.

Just a couple of comments + my review in #453.

@commoddity PTAL at #453 and edit/merge into this PR if it makes sense to you.

e2e/service_evm_test.go
commoddity: I noticed an issue with `eth_getTransactionReceipt` specifically returning the error only on websockets:
```bash
> {"jsonrpc":"2.0","method":"eth_getTransactionReceipt","params":["0xfeccd627b5b391d04fe45055873de3b2c0b4302d52e96bd41d5f0019a704165f"],"id":1}
< {"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"getReceipt error: seekInFiles(invIndex=unknown index,txNum=2786993528) but data before txNum=2900000000 not available"}}
```
This happens only for `eth` and only for websockets. By comparison, our fallback endpoint returns valid data for this exact `eth_getTransactionReceipt` request.

It seems to be an issue very specific to this method and only on websockets.
- The same `eth_getTransactionReceipt` request over HTTP succeeds.
- All other EVM methods, even archival ones, succeed over `eth` websockets.

For this reason I felt it was not worth hobbling our entire test suite for the purposes of this one method. That said, rather than removing it I have updated it to include ` TODO_INVESTIGATE` so this issue is not lost.

PR #463, opened 2025-09-25
Review of #455 (Adding legacy transforms for Grove Portal)

Olshansk:
Review of #455.

No blockers, but just a few nits/questions/suggestions related to naming.

### DIFF
--- a/portal-db/init/001_schema.sql
+++ b/portal-db/init/001_schema.sql
@@ -130,7 +130,7 @@ COMMENT ON COLUMN portal_users.portal_admin IS 'Whether user has admin privilege
 -- (portal_auth_type) a user is authenticated into the Portal by
 CREATE TABLE portal_user_auth (
     portal_user_auth_id SERIAL PRIMARY KEY,
-    portal_user_id VARCHAR(42),
+    portal_user_id VARCHAR(36),
     portal_auth_provider portal_auth_provider,
     portal_auth_type portal_auth_type,
     auth_provider_user_id VARCHAR(69),
--- a/portal-db/init/001_schema.sql
+++ b/portal-db/init/001_schema.sql
@@ -19,10 +19,12 @@ CREATE TYPE plan_interval AS ENUM ('day', 'month', 'year');
 -- Origin - Allow specific IP addresses or URLs
 CREATE TYPE allowlist_type AS ENUM ('service_id', 'contract', 'origin');
 
--- Add support for multiple auth providers offering different types
--- Easily extendable should additional authorization providers become necessary 
--- or requested
--- For legacy support, only adding auth0 and its relevant types
+-- Add support for multiple auth providers offering different types.
+-- Must be updated to extend authorization to other providers.
+--
+-- Grove's portal uses Auth0 for authentication as of 09/2025.
+--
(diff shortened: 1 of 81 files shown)


portal-db/init/001_schema.sql
```diff
@@ -78,7 +80,7 @@ COMMENT ON COLUMN portal_plans.plan_rate_limit_rps IS 'Rate limit in requests pe
 -- Portal Accounts can have many applications and many users. Only 1 user can be the OWNER.
 -- When a new user signs up in the Portal, they automatically generate a personal account.
 CREATE TABLE portal_accounts (
-    portal_account_id VARCHAR(36) PRIMARY KEY DEFAULT gen_random_uuid(),
+    portal_account_id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
```
fredteumer, line 83: can't do this -- we have integrations (google, stripe) so we can't change the data itself and therefore the data type.

portal-db/scripts/legacy-transform_up.sql
```diff
@@ -10,12 +10,12 @@ INSERT INTO public.portal_plans (
...
     plan_rate_limit_rps,
     plan_application_limit
-) 
+)
 VALUES
-    ('PLAN_FREE', 'Free tier with limited usage', 1000000, 'month', 0, 2),
-    ('PLAN_UNLIMITED', 'Unlimited Relays. Unlimited RPS.', 0, NULL, 0, 0),
-    ('PLAN_ENTERPRISE', 'Special case for Legacy Enterprise customers', 0, NULL, 0, 0),
-    ('PLAN_INTERNAL', 'Plan for internal accounts', 0, NULL, 0, 0);
+    ('PLAN_FREE', 'Free tier with limited monthly usage and rate limited.', 1e6, 'month', 100, 2),
```
fredteumer, line 15: this change is exceptionally nitpicky and pedantic and for an average grug DECREASES readability.

fredteumer: Also our Free tier has no rate limit currently, so it should still be 0 as the 2nd to last value

portal-db/scripts/legacy-transform_up.sql
```diff
@@ -10,12 +10,12 @@ INSERT INTO public.portal_plans (
...
+    ('PLAN_PAID', 'Paid tier with unlimited monthly usage (pay-as-you-go) and no rate limits.', 0, NULL, 0, 0),
+    ('PLAN_LEGACY', 'Legacy plan for enterprise customers', 0, NULL, 0, 0),
```
fredteumer, line 17: Again, I'm forced to ask why this is important? We reference this as `Enterprise` on the website (see pics in previous comment)

portal-db/scripts/legacy-transform_up.sql
```diff
@@ -10,12 +10,12 @@ INSERT INTO public.portal_plans (
...
+    ('PLAN_INTERNAL', 'Internal plan for development purposes only', 0, NULL, 0, 0);
```
fredteumer, line 18: We don't use this only for development, for example: Our hackathons account that we share with others. This is also used in production apps like status page. Disagree again here.

fredteumer (commented):
We could probably live without this PR given my comments. Bias is to shipping what is already in #455 and focusing on migrating/integration from here on out.

PR #466, opened 2025-10-02
[PostgREST] SDK Generation

Olshansk:
WIP


commoddity:
> Is this even necessary given that postgrest has compatible SDKs?

Given our preference for using fully typed code in our Go codebase - and especially avoiding usage of `interface{}`/`any` as much as possible - I think the autogenerated Go SDK provides a far superior API vs the https://github.com/supabase-community/postgrest-go client.

